PR #4, opened 2017-02-13
Add some benchmarking to docs

fredrikekre:
bench


docs/src/benchmarks.md
```diff
@@ -0,0 +1,33 @@
+# Benchmarks
+
+Here are some benchmark timings for tensors in 3 dimensions. For comparison
```
KristofferC: Should refer to the file where the benchmarks are generated.

fredrikekre:
Whats up with travis on nightly? Seems to be some stuff from #46 that is failing for some reason?

PR #2, opened 2017-02-10
Fixes for Julia 0.6

fredrikekre:
putting this branch here now instead

PR #5, opened 2017-02-24
[WIP] Generalize tensors multiplications

fredrikekre:


fredrikekre:
https://gist.github.com/fredrikekre/09422753ae60c1ab2c52db19439f0cae

KristofferC:
We could just copy StaticArrays eig function and say that it comes from there.

KristofferC:
AD got kinda rekt.

fredrikekre:
> We could just copy StaticArrays eig function and say that it comes from there.
Yea, could be a separate pr though.

I must have screwed something up with AD... I don't get why the Psi and S functions are allocating when used with dual numbers...

KristofferC:
Something is weird.. The `Psi` and `S` functions allocate even when the analytical function is called.

fredrikekre:
But only when called with dual numbers, right?

KristofferC:
No, Regressions are also only with Float32...

fredrikekre:
Yea, seems to be some kind of type-instability?

KristofferC:
```
julia> @code_warntype S(a)
Variables:
  #self#::#S
  C::Tensors.Tensor{2,2,Float32,4}

Body:
  begin 
      return $(Expr(:invoke, LambdaInfo for S(::Tensors.Tensor{2,2,Float32,4}, ::Float64, ::Float64), :(Main.S), :(C), :(Main.μ), :(Main.Kb)))
  end::Tensors.Tensor{2,2,Float64,4}
```

should not return a Float64?

KristofferC:
Or hmm, it should.. we just need to write the function more carefully

fredrikekre:
The power to `-1/3` promotes the `Float32` to `Float64`. But no reason it should allocate... Perhaps I messed up a conversion function actually.

KristofferC:
Reduced to subraction between tensors of different type:

```julia> C = rand(Tensor{2,2,Float64});

julia> CC = rand(Tensor{2,2,Float32});

julia> f(C,CC) = C - CC;

julia> @btime f($C,$CC);
  101.190 ns (4 allocations: 224 bytes)
```

fredrikekre:
Ehm...

```julia
julia> A = rand(Tensor{2, 3, Float64}); B = rand(Tensor{2, 3, Float32});

julia> @benchmark $A + $B
BenchmarkTools.Trial: 
  memory estimate:  400 bytes
  allocs estimate:  4
  --------------
  minimum time:     138.129 ns (0.00% GC)
  median time:      144.066 ns (0.00% GC)
  mean time:        162.921 ns (9.10% GC)
  maximum time:     1.489 μs (84.22% GC)
  --------------
  samples:          10000
  evals/sample:     873
  time tolerance:   5.00%
  memory tolerance: 1.00%
```

KristofferC:
LOL!

fredrikekre:
LOL at that sync

KristofferC:
Proof I'm the fastest debugger.

KristofferC:
Rekt by splatting. Try:

```jl
@inline Base.map{order, dim, T1, T2}(f, S1::AbstractTensor{order, dim, T1}, S2::AbstractTensor{order, dim, T2}) = ((SS1, SS2) = promote(S1, S2); _map(f, SS1, SS2))
```

fredrikekre:
Ah, thanks!

src/basic_operations.jl
```diff
@@ -0,0 +1,38 @@
...
+    quote
+        $(Expr(:meta, :inline))
+        @inbounds return $TensorType($expr)
+    end
+end
+
+@inline Base.map{order, dim, T1, T2}(f, S1::AbstractTensor{order, dim, T1}, S2::AbstractTensor{order, dim, T2}) = _map(f, promote(S1, S2)...)
+@inline Base.map{order, dim, T, N}(f, S1::Tensor{order, dim, T, N}, S2::Tensor{order, dim, T, N}) = _map(f ,S1, S2)
+@inline Base.map{order, dim, T, N}(f, S1::SymmetricTensor{order, dim, T, N}, S2::SymmetricTensor{order, dim, T, N}) = _map(f ,S1, S2)
+@generated function _map{T <: AllTensors}(f, S1::T, S2::T)
```
fredrikekre: btw, why do I need a `_map` here? I get overflow in line 27 if this method is also `map`. Two Tensors of exactly the same type, why doesn't that dispatch to this function?

KristofferC: works for me

KristofferC: using this file:

KristofferC: https://gist.github.com/KristofferC/0b1ab1a96a9c0f52e87d7ce0c79e5afa

KristofferC:
I pushed a commit to make the benchmarks keep the element type.

src/basic_operations.jl
fredrikekre: If you remove line 30 and 31 (which should be redundant 32)?
Thats how I had it first, that didnt work for me

fredrikekre:
Thanks. We should have benchmarks for basic ops like + and - too. Scary we wouldn't have caught this without the AD benchmarks ...

src/basic_operations.jl
fredrikekre: I guess it was some weird thing with the splatting again.

KristofferC:
Can we test this on 0.6 as well? It should work, right?

fredrikekre:
No. The stuff from #2 is not in here. (`typealias` etc...). Was hoping to merge this and then rebase #2.

KristofferC:
Makes sense

fredrikekre:
Re-running the benchmarks now.

fredrikekre:
https://gist.github.com/fredrikekre/dbf8c9f63a218df48af601a9121d722d
AD seems a bit sad still... The master run was without the type conversions though. But shouldn't make such a difference....

fredrikekre:
I have one more thing to try here -- but will do that tomorrow.

fredrikekre:
Guess which of these two are faster:

```julia
julia> Tensors.reducer(ex1, ex2)
:((((((((A[1] * B[1] + A[2] * B[2]) + A[3] * B[3]) + A[4] * B[4]) + A[5] * B[5]) + A[6] * B[6]) + A[7] * B[7]) + A[8] * B[8]) + A[9] * B[9])

julia> Tensors.reducer2(ex1, ex2)
:(A[9] * B[9] + (A[8] * B[8] + (A[7] * B[7] + (A[6] * B[6] + (A[5] * B[5] + (A[4] * B[4] + (A[3] * B[3] + (A[2] * B[2] + A[1] * B[1]))))))))
```

KristofferC:
I get exactly the same generated code.

fredrikekre:
The first case is about 20% faster here.

KristofferC:
```
julia> @btime f1($t, $t2)
  3.401 ns (0 allocations: 0 bytes)
2.283978437354851

julia> @btime f2($t, $t2)
  3.914 ns (0 allocations: 0 bytes)
2.283978437354851
```

But `code_native` says exactly the same code...

fredrikekre:
When that is used in double contraction:


```
julia> @code_native Tensors.dcontract1(A, B)
	.section	__TEXT,__text,regular,pure_instructions
Filename: tensor_products.jl
	pushq	%rbp
	movq	%rsp, %rbp
Source line: 35
	vmovsd	(%rdi), %xmm0           ## xmm0 = mem[0],zero
	vmulsd	(%rsi), %xmm0, %xmm0
	vmovsd	8(%rdi), %xmm1          ## xmm1 = mem[0],zero
	vmulsd	8(%rsi), %xmm1, %xmm1
	vaddsd	%xmm1, %xmm0, %xmm0
	vmovsd	16(%rdi), %xmm1         ## xmm1 = mem[0],zero
	vmulsd	16(%rsi), %xmm1, %xmm1
	vaddsd	%xmm1, %xmm0, %xmm0
	vmovsd	24(%rdi), %xmm1         ## xmm1 = mem[0],zero
	vmulsd	24(%rsi), %xmm1, %xmm1
	vaddsd	%xmm1, %xmm0, %xmm0
	vmovsd	32(%rdi), %xmm1         ## xmm1 = mem[0],zero
	vmulsd	32(%rsi), %xmm1, %xmm1
	vaddsd	%xmm1, %xmm0, %xmm0
	vmovsd	40(%rdi), %xmm1         ## xmm1 = mem[0],zero
	vmulsd	40(%rsi), %xmm1, %xmm1
	vaddsd	%xmm1, %xmm0, %xmm0
	vmovsd	48(%rdi), %xmm1         ## xmm1 = mem[0],zero
	vmulsd	48(%rsi), %xmm1, %xmm1
	vaddsd	%xmm1, %xmm0, %xmm0
	vmovsd	56(%rdi), %xmm1         ## xmm1 = mem[0],zero
	vmulsd	56(%rsi), %xmm1, %xmm1
	vaddsd	%xmm1, %xmm0, %xmm0
	vmovsd	64(%rdi), %xmm1         ## xmm1 = mem[0],zero
	vmulsd	64(%rsi), %xmm1, %xmm1
	vaddsd	%xmm1, %xmm0, %xmm0
	popq	%rbp
	retq
	nop

julia> @code_native Tensors.dcontract2(A, B)
	.section	__TEXT,__text,regular,pure_instructions
Filename: tensor_products.jl
	pushq	%rbp
	movq	%rsp, %rbp
Source line: 47
	vmovsd	64(%rdi), %xmm0         ## xmm0 = mem[0],zero
	vmulsd	64(%rsi), %xmm0, %xmm8
	vmovsd	56(%rdi), %xmm1         ## xmm1 = mem[0],zero
	vmulsd	56(%rsi), %xmm1, %xmm1
	vmovsd	48(%rdi), %xmm2         ## xmm2 = mem[0],zero
	vmulsd	48(%rsi), %xmm2, %xmm2
	vmovsd	40(%rdi), %xmm3         ## xmm3 = mem[0],zero
	vmulsd	40(%rsi), %xmm3, %xmm3
	vmovsd	32(%rdi), %xmm4         ## xmm4 = mem[0],zero
	vmulsd	32(%rsi), %xmm4, %xmm4
	vmovsd	24(%rdi), %xmm5         ## xmm5 = mem[0],zero
	vmulsd	24(%rsi), %xmm5, %xmm5
	vmovsd	16(%rdi), %xmm6         ## xmm6 = mem[0],zero
	vmulsd	16(%rsi), %xmm6, %xmm6
	vmovsd	(%rdi), %xmm7           ## xmm7 = mem[0],zero
	vmovsd	8(%rdi), %xmm0          ## xmm0 = mem[0],zero
	vmulsd	8(%rsi), %xmm0, %xmm0
	vmulsd	(%rsi), %xmm7, %xmm7
	vaddsd	%xmm7, %xmm0, %xmm0
	vaddsd	%xmm0, %xmm6, %xmm0
	vaddsd	%xmm0, %xmm5, %xmm0
	vaddsd	%xmm0, %xmm4, %xmm0
	vaddsd	%xmm0, %xmm3, %xmm0
	vaddsd	%xmm0, %xmm2, %xmm0
	vaddsd	%xmm0, %xmm1, %xmm0
	vaddsd	%xmm0, %xmm8, %xmm0
	popq	%rbp
	retq
	nop
```

KristofferC:
"kfischer [5:30 PM] 
First one interleaves the memory access with the arithmetic, which can help

[5:30]  
Though it shouldn’t make that big a difference
"

fredrikekre:
https://gist.github.com/fredrikekre/d06d1b1334453e17b138bbedf2d7f50c

KristofferC:
Looks good to me. We can maybe see if we can get the AD stuff back but it is not too big difference. Ok to merge by me.

fredrikekre:
I think the AD regression is due to the `dcontract` regression.

src/utilities.jl
```diff
@@ -23,22 +23,66 @@ function tensor_create{order, dim}(::Type{SymmetricTensor{order, dim}}, f)
...
     N = length(ex1)
-    ex = Expr(:call)
-    exn = Expr(:call, :*, ex1[1], ex2[1])
+    expr = :($(ex1[1]) * $(ex2[1]))
 
-    if N == 1 # return only the multiplication
-        return exn
+    for i in 2:N
+        expr = :($(expr) + $(ex1[i]) * $(ex2[i]))
+        # expr = :(muladd($(ex1[i]), $(ex2[i]), $expr)) # faster on some architectures!?
```
fredrikekre: This causes the regression for `dcontract` (which also is the reason for the AD benchmarks most likely). Before this PR we used `muladd` instructions but that turned out to be slower for `dot` and `norm`. It is really weird when `muladd` is faster or not. I will try to investigate this more and perhaps we can tailor the use of `muladd` a bit better.

fredrikekre:
https://gist.github.com/fredrikekre/d93d827bdeb4afd032081f779443bf2a

KristofferC:
Did u change something since last time?

fredrikekre:
https://gist.github.com/fredrikekre/6430b455923ee6305eab87a56d6386d2

fredrikekre:
Yes, see the two latest commits. AD is still sad :/ It is really weird, cause AD is faster for `dim` = 2...

KristofferC:
I think it is OK. Might be in the different cpu different result territory now.

fredrikekre:
I think that master run was just really good or something. It seems like the AD benchmarks are quite varying from run to run, but feels like this pr gets consistently worse result when run in the suite. In comparison, here is a run of the functions in the REPL: 

| function | PR | master |
|:---------|---:|-------:|
| `gradient Ψ - sym - Float64` | `388.282 ns` | `373.204` |
| `gradient Ψ - Float64` | `473.046 ns` | `470.214 ns` |
| `gradient S - sym - Float64` | `817.928 ns` | `833.867 ns` |
| `gradient S - Float64` | `1.544 μs` | `1.528 μs` |
| `hessian Ψ - sym - Float64` | `2.092 μs` | `2.104 μs` |
| `hessian Ψ - Float64` | `5.584 μs` | `5.559 μs` |
| `gradient Ψ - sym - Float32` | `346.815 ns` | `357.239 ns` |
| `gradient Ψ - Float32` | `429.487 ns` | `441.613 ns` |
| `gradient S - sym - Float32` | `690.819 ns` | `696.390 ns` |
| `gradient S - Float32` | `1.254 μs` | `1.253 μs` |
| `hessian Ψ - sym - Float32` | `1.959 μs` | `1.969 μs` |
| `hessian Ψ - Float32` | `4.460 μs` | `4.491 μs` |

And here is a new run (new master and pr run):
https://gist.github.com/fredrikekre/92efbe4b6dd7e8fa7bea078eca65d0ec

KristofferC:
Everything looks great from my point of view. Plz merge.

fredrikekre:
on my other computer:
https://gist.github.com/fredrikekre/3ad9bf05b45256dceaddabb861379ec1

fredrikekre:
These ones are really nice: 
| `["dot","dim 3 - order 2sym 2 - Float32"]` | 0.47 (5%) :white_check_mark: | 1.00 (1%)  |
| `["dot","dim 3 - order 2sym 2 - Float64"]` | 0.50 (5%) :white_check_mark: | 1.00 (1%)  |
| `["dot","dim 3 - order 2sym 2sym - Float64"]` | 0.77 (5%) :white_check_mark: | 1.00 (1%)  |

Weird we lost so much from promoting `SymmetricTensor` -> `Tensor` (which we did before)

src/Tensors.jl
```diff
@@ -104,245 +99,80 @@ Base.size{dim}(::FourthOrderTensor{dim}) = (dim, dim, dim, dim)
...
-        for dim in (1,2,3)
-            @eval @inline (::Type{Tensor{1, $dim}}){T <: Real}(t::SVector{$dim, T}) = Tensor{1, $dim, T, $dim}(t)
+        for dim in (1, 2, 3)
+            @eval @inline (::Type{Tensor{1, $dim}}){T <: Real}(t::NTuple{$dim, T}) = Tensor{1, $dim, T, $dim}(t)
         end
     end
 end
 # Special for Vec
 @inline (Tt::Type{Vec{dim}}){dim}(data) = Tensor{1, dim}(data)
+Base.convert{dim, T}(::Type{NTuple{dim, T}}, f::Function) = NTuple{dim, T}(ntuple(f, Val{dim}))
```
tkelman: can you avoid defining this? these are base types and can change the behavior of totally unrelated code

PR #8, opened 2017-02-27
fix evaluating one/zero in generator body

### LINKED ISSUE #7: `@generated` function generators don't follow the new rules for v0.6
The requirements for `@generated` and `@pure` functions have changed in Julia v0.6. For instance, AFAIK you can no longer call any function function from a function generator which may be specialized in the future, or you risk getting a world age error. My understanding of this is that function *generators* are compiled just once, and immediately as they are read in by the system, so they only have access to methods before their definition. Furthermore, any methods found of a later world age will result in a world age related error. I *think* the reason for this is that pure functions and generators do not populate back-edges in the dependency graph for automatic recompilation, but @vtjnash knows the exact details.

A specific example is [this line](https://github.com/KristofferC/Tensors.jl/blob/master/src/constructors.jl#L10) of *Tensors* where `one` and `zero` may fail on user defined types created *after* the user executes `using Tensors`. I haven't looked around for any more problems, but I thought I'd let you know after experience [this](https://github.com/JuliaArrays/StaticArrays.jl/issues/109) in *StaticArrays*.

KristofferC:
This should fix #7 cc @andyferris

### DIFF
--- a/src/constructors.jl
+++ b/src/constructors.jl
@@ -7,18 +7,20 @@ for TensorType in (SymmetricTensor, Tensor)
 
         @generated function Base.one{order, dim, T}(S::Type{$(TensorType){order, dim, T}})
             !(order in (2,4)) && throw(ArgumentError("`one` only defined for order 2 and 4"))
-            δ = (i,j) -> i == j ? :($(one(T))) : :($(zero(T)))
+            δ = (i,j) -> i == j ? :(o) : :(z)
             ReturnTensor = get_base(get_type(S))
             if order == 2
-                f = (i,j) -> δ(i,j)
+                f = (i,j) -> :($(δ(i,j)))
             elseif order == 4 && $TensorType == Tensor
-                f = (i,j,k,l) -> δ(i,k) * δ(j,l)
+                f = (i,j,k,l) -> :($(δ(i,k)) * $(δ(j,l)))
             else # order == 4 && TensorType == SymmetricTensor
-                f = (i,j,k,l) -> (δ(i,k) * δ(j,l) + δ(i,l) * δ(j,k)) / 2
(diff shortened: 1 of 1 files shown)


KristofferC:
Heh.. should just put u on review and u did it yourself 17 sec ago... Same wavelength brooo.

andyferris:
I didn't look over the all the codebase - is this the only place that this kind of thing happens? Does it solve the `using` vs `include` error you saw?

fredrikekre:
> Does it solve the using vs include error you saw?

No.

KristofferC:
Problem is here https://github.com/KristofferC/Tensors.jl/blob/49d600a05c07407f410bdf74ebe9eae444748929/src/constructors.jl#L42

I guess it is the same if the type is defined after that function.

andyferris:
Yeah, I saw that and it definitely raised some red flags in my mind, but I didn't exactly know what they were. It seemed a bit like the `similar_type `problem in *StaticArrays* (`similar_type` is the counterpart to `similar` for immutable arrays).

KristofferC:
But it is weird because the Tensor type is defined at that point. We don't expect anyone to subtype these types so we should have it easier than StaticArrays. Anyway, I'm not sure what the difference is between first `include` the module file and then `using` it ... Perhaps the `@generated` stuff runs but the types are still not really defined... I dunno. :/

andyferris:
Unlike most of Julia code, the order of generated functions w.r.t. other method definitions is important, even if they are simultaneously loaded from the same file. Maybe shuffling the order of stuff could help.

Not sure why `using` vs `include` should cause any difference there, however... Seems a mystery to me.

fredrikekre:
> Maybe shuffling the order of stuff could help.

Ref #9 

Thanks for the tip!

andyferris:
No worries. :)

PR #9, opened 2017-02-27
reshuffle stuff to make include work (world problem)

fredrikekre:


fredrikekre:
Putting all functions after the `get_type`.

PR #12, opened 2017-03-05
even better norm...

fredrikekre:
One easy simd implementation.

```julia
julia> @btime norm($A) # PR
  12.102 ns (0 allocations: 0 bytes)
5.381881970369471

julia> @btime norm($A) # master
  21.513 ns (0 allocations: 0 bytes)
5.38188197036947
```

### DIFF
--- a/src/math_ops.jl
+++ b/src/math_ops.jl
@@ -23,8 +23,14 @@ julia> norm(A)
 @inline Base.norm(v::Vec) = sqrt(dot(v, v))
 @inline Base.norm(S::SecondOrderTensor) = sqrt(dcontract(S, S))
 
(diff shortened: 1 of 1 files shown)


KristofferC:
This should work for all dimensions. Compiler should unroll by itself if loop is short enough.

KristofferC:
I wonder if `sqrt(sum(abs2, S))` is as fast?

fredrikekre:
No, `sqrt(mapreduce(abs2, +, S))` was much faster than `sqrt(sum(abs2, S))`

KristofferC:
It unrolls it automatically for `dim < 3`.

fredrikekre:
close in favor of #15

PR #13, opened 2017-03-06
Avoid defining convert from Function to NTuple

tkelman:
this is a pretty sketchy method - replace with some inner constructors instead


fredrikekre:
Thanks! I updated such that it works on both 0.5 and 0.6.

tkelman:
given you had some existing deprecations I wasn't sure how picky you'd be on that, but good call

fredrikekre:
For me it works on both 0.6 and 0.5 without deprecation warnings, but I think there are some dependencies that throw some warnings. Did you get warnings from this package? :confused:

tkelman:
I was just looking at the travis logs, and didn't check too closely exactly where they came from

fredrikekre:
Travis seems to have timed out. Thanks for the fix!

PR #14, opened 2017-03-06
benchmark with dual numbers

KristofferC:


KristofferC:
https://gist.github.com/KristofferC/ae89cadc45202bbb53fba54649e283a8

Something is weird with the timings of the constructors. They are fast in the REPL.

fredrikekre:
Think it is something weird when we interpolate the types, see this:
```julia
julia> T = Tensor{4, 3};

julia> @btime one($T);
  140.230 ns (1 allocation: 672 bytes)

julia> @btime one(Tensor{4, 3});
  12.691 ns (0 allocations: 0 bytes)
```

fredrikekre:
I am adding some more benchmarks, so I will merge this such that there wont be conflicts.

PR #16, opened 2017-03-10
Introduce a dummy type F64 to test with

fredrikekre:
This allows the "normal" methods to still be tested after the simd-forcing methods is introduced.

### DIFF
--- a/test/F64.jl
+++ b/test/F64.jl
@@ -1,0 +1,31 @@
+# dummy type wrapping a Float64 used in tests
+immutable F64 <: AbstractFloat
+    x::Float64
+end
+
+# operations
+for op in (:+, :-)
+    @eval Base.$op(a::F64) = F64($op(a.x))
+end
+for op in (:+, :-, :*, :/)
+    @eval Base.$op(a::F64, b::F64) = F64($op(a.x, b.x))
+end
+for op in(:zero, :one)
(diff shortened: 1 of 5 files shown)


fredrikekre:
It's pretty cool that such a limited number of operations are needed on this type to make all our tensor functions to work. :tada:

fredrikekre:
> It's not like people will use types not defined in base with Tensors so It wont matter that much with the world age stuff

Famous last words. Nice that we caught this at least :) I think the package is world age safe now.

KristofferC:
Nice!

test/F64.jl
```diff
@@ -0,0 +1,32 @@
...
+for op in (:+, :-)
+    @eval Base.$op(a::F64) = F64($op(a.x))
+end
+for op in (:+, :-, :*, :/)
+    @eval Base.$op(a::F64, b::F64) = F64($op(a.x, b.x))
+end
+for op in(:zero, :one)
+    @eval Base.$op(::Type{F64}) = F64($op(Float64))
+end
+Base.rand(rng::AbstractRNG, F64) = F64(rand())
```
tkelman: is this supposed to say `::F64` or `::Type{F64}` ?

fredrikekre: Probably `::Type{F64}`

PR #17, opened 2017-03-11
Explicit simd for ad

KristofferC:
Unfortunately no way to add this without copy pasting a bunch of code from ForwardDiff.

Benchmarks (on my laptop): https://gist.github.com/KristofferC/9b62f7c130cc4eef256092c363c8a01c

### DIFF
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -1,0 +1,3 @@
+ignore:
+  - "src/forwarddiff/dual.jl"
+  - "src/forwarddiff/partials.jl"
--- a/REQUIRE
+++ b/REQUIRE
@@ -2,3 +2,6 @@ julia 0.5
 ForwardDiff
 SIMD
 Compat 0.19.0
+Calculus 0.2.0
+NaNMath 0.2.2
+SpecialFunctions 0.1.0
--- a/benchmark/benchmark_ad.jl
+++ b/benchmark/benchmark_ad.jl
@@ -1,9 +1,8 @@
-import ForwardDiff: valtype
 const ∇ = Tensors.gradient
 const Δ = Tensors.hessian
 
 function Ψ(C, μ, Kb)
-    T = ForwardDiff.valtype(eltype(C))
+    T = Tensors.valtype(eltype(C))
     detC = det(C)
     J = sqrt(detC)
     Ĉ = detC^(T(-1 / 3)) * C
@@ -11,7 +10,7 @@ function Ψ(C, μ, Kb)
 end
 
 function S(C, μ, Kb)
-    T = ForwardDiff.valtype(eltype(C))
+    T = Tensors.valtype(eltype(C))
     I = one(C)
     J = sqrt(det(C))
     invC = inv(C)
--- a/benchmark/runbenchmarks.jl
+++ b/benchmark/runbenchmarks.jl
@@ -1,5 +1,6 @@
 using Tensors
 using BenchmarkTools
+using ForwardDiff
 using JLD
 
 include("generate_report.jl")
--- a/benchmark/runbenchmarks.jl
+++ b/benchmark/runbenchmarks.jl
@@ -1,6 +1,5 @@
 using Tensors
 using BenchmarkTools
-using ForwardDiff
 using JLD
 
 include("generate_report.jl")
@@ -10,7 +9,7 @@ const ALL_DIMENSIONS = true
 const MIXED_SYM_NONSYM = true
 const MIXED_ELTYPES = true
 
-const dT = ForwardDiff.Dual{4, Float64}
+const dT = Tensors.Dual{4, Float64}
 
 function create_tensors()
     tensor_dict = Dict{Tuple{Int, Int, DataType}, AbstractTensor}()
--- a/benchmark/runbenchmarks.jl
+++ b/benchmark/runbenchmarks.jl
@@ -1,6 +1,5 @@
 using Tensors
 using BenchmarkTools
-using ForwardDiff
 using JLD
 
 include("generate_report.jl")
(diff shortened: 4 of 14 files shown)


src/partials.jl
```diff
@@ -0,0 +1,246 @@
...
+@generated function minus_tuple{N}(tup::NTuple{N})
+    return tupexpr(i -> :(-tup[$i]), N)
+end
+
+@generated function mul_tuples{N}(a::NTuple{N}, b::NTuple{N}, afactor, bfactor)
+    return tupexpr(i -> :((afactor * a[$i]) + (bfactor * b[$i])), N)
+end
+
+########
+# SIMD #
```
KristofferC: This and below is the only thing we actually need.

KristofferC:
SIMD O3 vs SIMD: https://gist.github.com/KristofferC/95f7e2c4ca0dc70353858cce06b27ea4
SIMD O3 vs O3: https://gist.github.com/KristofferC/ff54ad0193b62211805b8df1de2fc877
SIMD vs O3: https://gist.github.com/KristofferC/0f8cd24b15a5c1c163d2c3447df09a85

Basically, O3 beats SIMD (with default optimization). SIMD with O3 beats both O3 and SIMD without O3.

KristofferC:
A conclusion here is that it is not good to measure O3 vs SIMD. We should measure SIMD O3 vs O3.

fredrikekre:
Nice! What parts of the code did you add yourself? Just the SIMD section in partials.jl?
This seems like an improvement for the ForwardDiff package too? Perhaps it could be upstreamed?

KristofferC:
Only the SIMD stuff. We are just lucky that we do not hit the bad cases (length 7 for example). This is why this can't really be upstreamed.

src/partials.jl
```diff
@@ -0,0 +1,246 @@
...
+
+@inline function scale_tuple{N, T <: SIMDTypes}(tup::NTuple{N, T}, x)
+    return totuple(tosimd(tup) * x)
+end
+
+@inline function div_tuple_by_scalar{N, T <: SIMDTypes}(tup::NTuple{N, T}, x)
+    return totuple(tosimd(tup) / x)
+end
+
+@inline function add_tuples{N, T1 <: SIMDTypes, T2 <: SIMDTypes}(a::NTuple{N, T1}, b::NTuple{N, T2})
```
fredrikekre: Need to limit this and `sub_tuples` below to the same type `T`:

```julia
julia> v1 = Vec{4, Float64}((rand(4)...));

julia> v2 = Vec{4, Float32}((rand(Float32, 4)...));

julia> v1 + v2
4-element Array{Float64,1}:
 1.26787
 1.376  
 1.27875
 1.11322
```

KristofferC: I don't see what is bad about that?

KristofferC:
Rip code coverage...

src/partials.jl
fredrikekre: It turns into an array?
```julia
julia> @which v1 + v2
+{R,S}(A::AbstractArray{R,N<:Any}, B::AbstractArray{S,N<:Any}) at arraymath.jl:63
```

KristofferC: Ooh!

fredrikekre: Could fix that in SIMD.jl

KristofferC:
Can you try:

```
import Tensors.Dual
using BenchmarkTools
x, y, z = Dual(rand(9)...) , Dual(rand(9)...), Dual(rand(9)...);
fma2(x,y,z) = x*y + z
@benchmark muladd($x, $y, $z)
@benchmark fma2($x, $y, $z)
```

on your fancy new ass CPU?

fredrikekre:
On this branch?

fredrikekre:
```julia
julia> @benchmark muladd($x, $y, $z)
BenchmarkTools.Trial: 
  memory estimate:  0 bytes
  allocs estimate:  0
  --------------
  minimum time:     2.406 ns (0.00% GC)
  median time:      2.417 ns (0.00% GC)
  mean time:        2.441 ns (0.00% GC)
  maximum time:     18.330 ns (0.00% GC)
  --------------
  samples:          10000
  evals/sample:     1000

julia> @benchmark fma2($x, $y, $z)
BenchmarkTools.Trial: 
  memory estimate:  0 bytes
  allocs estimate:  0
  --------------
  minimum time:     7.638 ns (0.00% GC)
  median time:      7.641 ns (0.00% GC)
  mean time:        7.750 ns (0.00% GC)
  maximum time:     25.681 ns (0.00% GC)
  --------------
  samples:          10000
  evals/sample:     999
```

KristofferC:
Cool. When u have time can u also benchmark the AD stuff and compare it to master.

fredrikekre:
https://gist.github.com/fredrikekre/4a4887eaccd617b79337583c838a616f

KristofferC:
So this PR is crap?

fredrikekre:
The analytical benchmarks in there should not be affected by this PR, right? Even they show bad results...

> So this PR is crap?

Perhaps simd should be limited to dim = 3 atleast?

fredrikekre:
By the way, we are not benchmarking with symmetric tensors due to https://github.com/KristofferC/Tensors.jl/blob/7728952ee6fdcc834c61205f6ac628cee556e68c/benchmark/benchmark_ad.jl#L35 not sure if intended?

fredrikekre:
Btw, seems to be commits from #19 on this branch...

fredrikekre:
New run:
https://gist.github.com/fredrikekre/8569d885c34556120e666adb4f0ac94a
Idk what was up with last run.

fredrikekre:
This PR does nothing now?

KristofferC:
Must have messed up a rebase...

KristofferC:
Some pretty nasty compilation time changes... :(

KristofferC:
Let's try this

KristofferC:
Not inlining some functions when using dim = 3 and dual numbers helped a bit. Just need to check that benchmarks were not affected.

KristofferC:
Turns out I didn't need to import all of ForwardDiff because we specify stricter version of the functions lol...

fredrikekre:
This PR is a roller coaster :roller_coaster: . I feel like every second commit deleted thousand lines and the next one adds them back xD

KristofferC:
I dunno wtf is going on with these rebases...

fredrikekre:
Which side is this PR?

KristofferC:
Right

fredrikekre:
This is also interesting: with:

```julia
julia> A = rand(Tensor{2, 3}); B = symmetric(A);
```

I get on master:
```julia
julia> @btime dot($A, $B);
  7.120 ns (0 allocations: 0 bytes)
```

and with this PR (which changed to muladd):

```julia
julia> @btime dot($A, $B);
  4.734 ns (0 allocations: 0 bytes)
```

 I am sure it was slower with `muladd` for `dot`, thats why it wasn't enabled...

KristofferC:
SIMD.jl is now used upstream in ForwardDiff so this can be closed.

PR #18, opened 2017-03-12
clean up benchmarks

fredrikekre:
- clean up
- add some missing cases
- "pretty printing"

https://gist.github.com/fredrikekre/6fd9e2cededb71fcce907c0985757a3c

### DIFF
--- a/benchmark/benchmark_ad.jl
+++ b/benchmark/benchmark_ad.jl
@@ -27,7 +27,7 @@ S(C) = S(C, μ, Kb)
 SUITE["gradient"] = BenchmarkGroup(["ad"])
 SUITE["hessian"] = BenchmarkGroup(["ad"])
 
-for dim in (1,2,3)
+for dim in (ALL_DIMENSIONS ? (1,2,3) : (3,))
     for T in (Float64, Float32)
         V2 = tensor_dict[(dim, 2, T)]
         V2 = V2' ⋅ V2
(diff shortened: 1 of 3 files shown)


KristofferC:
Just a thing I noticed. This:

```jl
function otimes2{dim}(a::Tensor{2,dim}, b::Tensor{2, dim})
    Tensor{4,dim}(@inline function(i,j,k,l) @inbounds v = a[i,j]*b[k,l]; v end)
end
```

produces the same code as our `otimes`.

fredrikekre:
Thats interesting. But:
```julia
julia> @btime otimes($A, $A);
  4.549 ns (0 allocations: 0 bytes)

julia> @btime otimes2($A, $A);
  11.208 ns (0 allocations: 0 bytes)
```

KristofferC:
I saw that too... makes no sense

fredrikekre:
See my edit above

KristofferC:
Ah!

fredrikekre:
So this turns out to be quite a powerful function then:
https://github.com/KristofferC/Tensors.jl/blob/master/src/constructors.jl#L4-L17

KristofferC:
Yes.

fredrikekre:
Can perhaps simplify a lot with that method?
```julia
@inline function zero2{T}(::Type{Tensor{2, 3, T}})
    Tensor{2, 3}(@inline function(i, j) zero(T); end)
end
```

KristofferC:
Yes, I think so too!

KristofferC:
```
@inline function Base.transpose{dim}(S::Tensor{2, dim})
    Tensor{2, dim}(@inline function(i, j) @inbounds v = S[j,i]; v; end)
end
```

fredrikekre:
`+, -, *, /` as well. I guess it can't be done when there are reductions involved, like `dot`

PR #19, opened 2017-03-12
Simplify some functions

KristofferC:
TODO:

- [ ] fix more functions
- [ ] benchmark

### DIFF
--- a/src/constructors.jl
+++ b/src/constructors.jl
@@ -45,7 +45,6 @@ end
     M = n_components(SymmetricTensor{order,dim})
     expM = Expr(:tuple, [:(data[$i]) for i in 1:M]...)
     return quote
-        # println("BAD DISPATCH")
         L = length(data)
         if L != $N && L != $M
             throw(ArgumentError("wrong number of vector elements, expected $($N) or $($M), got $L"))
--- a/src/constructors.jl
+++ b/src/constructors.jl
@@ -16,7 +16,6 @@
     end
 end
 
-# Tensor from AbstractArray
 @generated function (::Type{Tensor{order, dim}}){order, dim}(data::AbstractArray)
     N = n_components(Tensor{order,dim})
     exp = Expr(:tuple, [:(data[$i]) for i in 1:N]...)
@@ -24,10 +23,22 @@ end
         if length(data) != $N
             throw(ArgumentError("wrong number of elements, expected $($N), got $(length(data))"))
         end
-        # println("BAD DISPATCH")
         Tensor{order, dim}($exp)
     end
 end
+
+@inline function apply_all{order, dim}(S::Union{Tensor{order, dim}, SymmetricTensor{order, dim}}, f::Function)
+    apply_all(get_base(typeof(S)), f)
+end
+
+# Tensor from AbstractArray
+function (T::Type{Tensor{order, dim}}){order, dim}(data::AbstractArray)
+    N = n_components(T)
+    length(data) != n_components(T) && throw(ArgumentError("wrong number of elements, expected $N, got $(length(data))"))
+    return apply_all(T, @inline function(i) @inboundsret data[i]; end)
+end
+
+
 
 # SymmetricTensor from AbstractArray
 @generated function (::Type{SymmetricTensor{order, dim}}){order, dim}(data::AbstractArray)
@@ -91,16 +102,8 @@ end
 @eval @inline Base.$op(t::AllTensors) = $op(typeof(t))
 end
 
-@generated function Base.fill{T <: AbstractTensor}(el::Union{Number, Function}, S::Type{T})
-    TensorType = get_base(get_type(S))
-    N = n_components(TensorType)
-    ele = el <: Number ? :(el) : :(el())
-    expr = Expr(:tuple, [ele for i in 1:N]...)
-    quote
-        $(Expr(:meta, :inline))
-        @inbounds return $TensorType($expr)
-    end
-end
+Base.fill{T <: AbstractTensor}(el::Number, S::Type{T}) = apply_all(get_base(T), i -> el)
+Base.fill{T <: AbstractTensor}(f::Function, S::Type{T}) = apply_all(get_base(T), i -> f())
 
 # Array with zero/ones
 @inline Base.zeros{T <: AbstractTensor}(::Type{T}, dims...) = fill(zero(T), dims...)
--- a/src/constructors.jl
+++ b/src/constructors.jl
@@ -16,14 +16,13 @@
     end
 end
 
-@generated function (::Type{Tensor{order, dim}}){order, dim}(data::AbstractArray)
-    N = n_components(Tensor{order,dim})
-    exp = Expr(:tuple, [:(data[$i]) for i in 1:N]...)
-    return quote
-        if length(data) != $N
-            throw(ArgumentError("wrong number of elements, expected $($N), got $(length(data))"))
-        end
-        Tensor{order, dim}($exp)
+# Applies the function f to all indices f(1), f(2), ... f(n_independent_components)
+@generated function apply_all{order, dim}(S::Union{Type{Tensor{order, dim}}, Type{SymmetricTensor{order, dim}}}, f::Function)
+    TensorType = get_base(get_type(S))
+    exp = tensor_create_linear(TensorType, (i) -> :(f($i)))
+    quote
+        $(Expr(:meta, :inline))
+        @inbounds return $TensorType($exp)
     end
 end
 
@@ -37,7 +36,6 @@ function (T::Type{Tensor{order, dim}}){order, dim}(data::AbstractArray)
     length(data) != n_components(T) && throw(ArgumentError("wrong number of elements, expected $N, got $(length(data))"))
     return apply_all(T, @inline function(i) @inboundsret data[i]; end)
 end
-
 
 
 # SymmetricTensor from AbstractArray
--- a/src/math_ops.jl
+++ b/src/math_ops.jl
@@ -263,17 +263,15 @@ julia> trace(dev(A))
 0.0
 ```
 """
-@generated function dev{dim}(S::SecondOrderTensor{dim})
-    Tt = get_base(S)
-    idx(i,j) = compute_index(Tt, i, j)
-    f = (i,j) -> i == j ? :((get_data(S)[$(idx(i,j))] - tr/3)) :
-                           :(get_data(S)[$(idx(i,j))])
-    exp = tensor_create(Tt, f)
-    return quote
-        $(Expr(:meta, :inline))
-        tr = trace(S)
-        $Tt($exp)
-    end
+@inline function dev(S::SecondOrderTensor)
+    Tt = get_base(typeof(S))
+    tr = trace(S) / 3
+    Tt(
+        @inline function(i, j)
+            @inbounds  v = i == j ? S[i,j] - tr : S[i,j]
+            v
+        end
+    )
 end
 
 # http://inside.mines.edu/fs_home/gmurray/ArbitraryAxisRotation/
--- a/src/math_ops.jl
+++ b/src/math_ops.jl
@@ -58,28 +58,12 @@ julia> det(A)
 -0.1005427219925894
 ```
 """
-@generated function Base.det{dim}(t::SecondOrderTensor{dim})
-    idx(i,j) = compute_index(get_base(t), i, j)
-    if dim == 1
-        return quote
-            $(Expr(:meta, :inline))
-            get_data(t)[$(idx(1,1))]
-        end
-    elseif dim == 2
-        return quote
-            $(Expr(:meta, :inline))
-            v = get_data(t)
-            v[$(idx(1,1))] * v[$(idx(2,2))] - v[$(idx(1,2))] * v[$(idx(2,1))]
-        end
-    else # dim == 3
-        return quote
-            $(Expr(:meta, :inline))
-            v = get_data(t)
-            (v[$(idx(1,1))]*(v[$(idx(2,2))]*v[$(idx(3,3))]-v[$(idx(2,3))]*v[$(idx(3,2))]) -
-             v[$(idx(1,2))]*(v[$(idx(2,1))]*v[$(idx(3,3))]-v[$(idx(2,3))]*v[$(idx(3,1))]) +
-             v[$(idx(1,3))]*(v[$(idx(2,1))]*v[$(idx(3,2))]-v[$(idx(2,2))]*v[$(idx(3,1))]))
-        end
-    end
+@inline Base.det(t::SecondOrderTensor{1}) = @inboundsret t[1,1]
+@inline Base.det(t::SecondOrderTensor{2}) = @inboundsret (t[1,1] * t[2,2] - t[1,2] * t[2,1])
+@inline function Base.det(t::SecondOrderTensor{3})
+    @inboundsret (t[1,1] * (t[2,2]*t[3,3] - t[2,3]*t[3,2]) -
+                  t[1,2] * (t[2,1]*t[3,3] - t[2,3]*t[3,1]) +
+                  t[1,3] * (t[2,1]*t[3,2] - t[2,2]*t[3,1]))
 end
 
 """
--- a/src/tensor_products.jl
+++ b/src/tensor_products.jl
@@ -313,14 +313,6 @@ julia> a × b
   0.116354
 ```
 """
-function Base.cross{T}(u::Vec{3, T}, v::Vec{3, T})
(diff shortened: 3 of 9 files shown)


src/tensor_products.jl
```diff
@@ -131,19 +131,9 @@ julia> A ⊗ B
...
-    for l in 1:dim, k in 1:dim, j in 1:dim, i in 1:dim
-        push!(exps.args, :(get_data(S1)[$(idxS1(i, j))] * get_data(S2)[$(idxS2(k, l))]))
-    end
-    expr = remove_duplicates(TensorType, exps)
-    quote
-        $(Expr(:meta, :inline))
-        @inbounds return $TensorType($expr)
-    end
+@inline function otimes{dim}(S1::SecondOrderTensor{dim}, S2::SecondOrderTensor{dim})
+    TensorType = getreturntype(otimes, get_base(typeof(S1)), get_base(typeof(S2)))
```
fredrikekre: This will only be fast if `getreturntype` is `@pure`

KristofferC:
We could perhaps have some "reduce" function to simplify other things

src/math_ops.jl
```diff
@@ -263,17 +263,18 @@ julia> trace(dev(A))
...
-    return quote
-        $(Expr(:meta, :inline))
-        tr = trace(S)
-        $Tt($exp)
-    end
+@inline function dev(S::SecondOrderTensor)
+    Tt = get_base(typeof(S))
+    tr = trace(S) / 3
+    Tt(
+        @inline function(i, j)
```
fredrikekre: This function will return different eltype for `Int` tensors.

fredrikekre:
```julia
julia> Tensor{2, 3}((i, j)-> i == j ? 1 : 1.0)
ERROR: MethodError: Cannot `convert` an object of type Tuple{Int64,Float64,Float64,Float64,Int64,Float64,Float64,Float64,Int64} to an object of type Tensors.Tensor{2,3,T<:Real,M}
This may have arisen from a call to the constructor Tensors.Tensor{2,3,T<:Real,M}(...),
since type constructors fall back to convert methods.
 in Tensors.Tensor{2,3,T<:Real,M}(::Tuple{Int64,Float64,Float64,Float64,Int64,Float64,Float64,Float64,Int64}) at ./sysimg.jl:53
 in macro expansion at /home/fredrik/Dropbox/Programming/Tensors.jl/src/constructors.jl:15 [inlined]
 in Tensors.Tensor{2,3,T<:Real,M}(::##1#2) at /home/fredrik/Dropbox/Programming/Tensors.jl/src/constructors.jl:5
```

KristofferC: Yep. I tried:

```
@inline function dev(S::SecondOrderTensor)
    Tt = get_base(typeof(S))
    tr = trace(S) / 3
    T = typeof(one(eltype(S)) / 3)
    Tt(
        @inline function(i, j)
            @inbounds if i == j
                return T(S[i,j] - tr)
            else
                return T(S[i,j])
            end
        end
    )
end
```

but it is slow for some reason.

fredrikekre: Could do this:
```julia
@inline function dev{dim, T}(S::SecondOrderTensor{dim, T})
    Tt = get_base(typeof(S))
    tr = trace(S) / 3
    R = promote_type(T, typeof(tr))
    Tt(
        @inline function(i, j)
            @inbounds if i == j
                return R(S[i,j] - tr)
            else
                return R(S[i,j])
            end
        end
    )
end
```

fredrikekre: Edit, yea its really slow.

KristofferC: Looks like `Tt` is not inferred..

fredrikekre: It doesn't work on master either so
```julia
julia> A = rand(Tensor{2, 3, Int});

julia> dev(A)
ERROR: MethodError: Cannot `convert` an object of type Tuple{Float64,Int64,Int64,Int64,Float64,Int64,Int64,Int64,Float64} to an object of type Tensors.Tensor{2,3,T<:Real,M}
This may have arisen from a call to the constructor Tensors.Tensor{2,3,T<:Real,M}(...),
since type constructors fall back to convert methods.
 in macro expansion at /home/fredrik/Dropbox/Programming/Tensors.jl/src/math_ops.jl:275 [inlined]
 in dev(::Tensors.Tensor{2,3,Int64,9}) at /home/fredrik/Dropbox/Programming/Tensors.jl/src/math_ops.jl:267
```

src/transpose.jl
```diff
@@ -63,15 +49,8 @@ majortranspose(::FourthOrderTensor)
...
-@generated function majortranspose{dim}(S::FourthOrderTensor{dim})
-    expr = Expr(:tuple)
-    for l in 1:dim, k in 1:dim, j in 1:dim, i in 1:dim
-        push!(expr.args, :(get_data(S)[$(compute_index(get_base(S), k, l, i, j))]))
-    end
-    return quote
-        $(Expr(:meta, :inline))
-        @inbounds return Tensor{4, dim}($expr)
-    end
+@inline function majortranspose{dim}(S::Tensor{4, dim})
```
fredrikekre: This method should have `S::FourthOrderTensor{dim}` as input but always return `Tensor{4, dim}` so just change the input type signature.

KristofferC:
Did some more functions.. Using the function stuff is really powerful.

src/math_ops.jl
KristofferC: I put it back since it fails on master

KristofferC:
Jesus we have a lot of benchmarks...

src/math_ops.jl
fredrikekre: We could perhaps limit eltype for this method to `AbstractFloat`?

src/transpose.jl
```diff
@@ -24,15 +24,8 @@ julia> A'
...
-    expr = Expr(:tuple)
-    for j in 1:dim, i in 1:dim
-        push!(expr.args, :(get_data(S)[$(compute_index(Tensor{2, dim}, j, i))]))
-    end
-    quote
-        $(Expr(:meta, :inline))
-        @inbounds return Tensor{2, dim}($expr)
-    end
+@inline function Base.transpose{dim}(S::Tensor{2, dim})
+    Tensor{2, dim}(@inline function(i, j) @inbounds v = S[j,i]; v; end)
```
fredrikekre: `@inboundsret` ?

fredrikekre:
This is great!

> Jesus we have a lot of benchmarks...

Yea, better safe than sorry :p

KristofferC:
Benchmarks here: https://gist.github.com/KristofferC/8f423c6d4b2c1f614b3ff7f03fa7c394

I did not find any difference in the generated code for anything that looked like it could be real regressions.

fredrikekre:
I tested some cases too, didn't see any difference. Benchmarks really are weird...

KristofferC:
Pfft 36 min on Travis 0.6. Is this due to a change in Base or this PR?

fredrikekre:
Locally on v06:
```
278.792679 seconds (1.01 G allocations: 21.509 GiB, 2.43% gc time) # master
479.304849 seconds (3.57 G allocations: 63.570 GiB, 3.17% gc time) # PR
```

KristofferC:
Yuck... hmm.

fredrikekre:
https://github.com/JuliaLang/julia/issues/18077 ?

KristofferC:
```
 Section               ncalls     time   %tot     avg     alloc   %tot      avg
 ──────────────────────────────────────────────────────────────────────────────
 tensor ops                 1     923s  57.6%    923s    131GiB  61.9%   131GiB
   symmetric/skew-s...      9     665s  41.5%   73.9s    101GiB  47.5%  11.2GiB
   transpose                9     159s  9.93%   17.7s   22.4GiB  10.5%  2.49GiB
   outer product            9    73.6s  4.59%   8.18s   7.26GiB  3.42%   826MiB
   double contraction       9    21.9s  1.37%   2.43s    779MiB  0.36%  86.6MiB
   dot products             9    1.82s  0.11%   202ms   55.1MiB  0.03%  6.12MiB
   special                  9    1.12s  0.07%   124ms   36.7MiB  0.02%  4.08MiB
   cross product            9    186ms  0.01%  20.7ms   3.96MiB  0.00%   450KiB
   rotation                 9    362μs  0.00%  40.2μs   40.6KiB  0.00%  4.52KiB
 constrct Arr               1     281s  17.5%    281s   41.8GiB  19.7%  41.8GiB
 AD                         1     149s  9.30%    149s   11.9GiB  5.61%  11.9GiB
   dim 3                    1     129s  8.06%    129s   11.1GiB  5.25%  11.1GiB
     hessian                2    70.2s  4.38%   35.1s   6.79GiB  3.20%  3.39GiB
     2nd tensor grad        2    23.3s  1.45%   11.6s   1.60GiB  0.75%   817MiB
     scalar grad            2    2.05s  0.13%   1.03s    113MiB  0.05%  56.4MiB
   dim 2                    1    14.3s  0.89%   14.3s    625MiB  0.29%   625MiB
     hessian                2    6.39s  0.40%   3.19s    293MiB  0.13%   147MiB
     2nd tensor grad        2    3.39s  0.21%   1.69s    132MiB  0.06%  65.8MiB
     scalar grad            2    1.15s  0.07%   573ms   41.2MiB  0.02%  20.6MiB
   dim 1                    1    5.53s  0.34%   5.53s    163MiB  0.07%   163MiB
     hessian                2    2.65s  0.17%   1.33s   81.9MiB  0.04%  40.9MiB
     2nd tensor grad        2    1.09s  0.07%   543ms   24.8MiB  0.01%  12.4MiB
     scalar grad            2    549ms  0.03%   274ms   16.0MiB  0.01%  8.00MiB
 indexing                   1     136s  8.47%    136s   20.9GiB  9.84%  20.9GiB
 tensor identities          1    44.9s  2.80%   44.9s   3.85GiB  1.81%  3.85GiB
 constructors               1    22.8s  1.42%   22.8s    700MiB  0.32%   700MiB
 simple math                1    17.1s  1.07%   17.1s    724MiB  0.33%   724MiB
 norm, trace, det, ...      1    13.0s  0.81%   13.0s    484MiB  0.22%   484MiB
 promotion/conversion       1    7.32s  0.46%   7.32s    301MiB  0.14%   301MiB
 base vectors               1    3.66s  0.23%   3.66s    137MiB  0.06%   137MiB
 diagm, one                 1    3.41s  0.21%   3.41s   93.9MiB  0.04%  93.9MiB
 constrct func              1    2.64s  0.16%   2.64s   96.5MiB  0.04%  96.5MiB
 exceptions                 1   2.14ms  0.00%  2.14ms   7.89KiB  0.00%  7.89KiB
 ──────────────────────────────────────────────────────────────────────────────
```
symmetric /skew symmetric  600 seconds!, one master, 3.52 seconds.

fredrikekre:
Seems to be extremely hard on inference with this method...

KristofferC:
I don't get it though.. on 0.5.1:

```
<< using Tensors

<< a  = rand(Tensor{4,3});

<< @time symmetric(a);
 51.641367 seconds (68.55 M allocations: 1.077 GB, 0.48% gc time)
```

so why is it fast when doing the tests?

fredrikekre:
That time is just ridiculous LOL

fredrikekre:
I tried typeasserting the functions but didn’t make a difference.

KristofferC:
This is fast though:

```jl
immutable Tensor{order, dim, T <: Real, M}
    data::NTuple{M, T}
end

@generated function gg{order, dim}((S::Type{Tensor{order, dim}}), f::Function)
    exp = Expr(:tuple, [:(f($i,$j,$k,$l)) for i=1:dim, j=1:dim, k = 1:dim, l = 1:dim]...)
    return quote
        return $exp
    end
end

gg(Tensor{4,3}, (i,j,k,l) -> i+j+k+l)
```

KristofferC:
With https://github.com/JuliaLang/julia/pull/21085 tests on this branch are slightly faster than on master.

KristofferC:
It seems that we can generally avoid using the `compute_index` stuff to avoid the indexing because the compiler does it by itself.
For example:

```jl
@generated function dcontract_muladd{dim}(S1::Tensor{2, dim}, S2::Tensor{2, dim})
    ex1 = Expr[:(S1[$i, $j]) for i in 1:dim, j in 1:dim][:]
    ex2 = Expr[:(S2[$i, $j]) for i in 1:dim, j in 1:dim][:]
    exp = reducer(ex1, ex2, true)
    return quote
        $(Expr(:meta, :inline))
        @inbounds return $exp
    end
end
```

produces just as good code as the one with `compute_index`.

fredrikekre:
Yea, but for `SymmetricTensor` we can not do that

KristofferC:
Why not?

fredrikekre:
the `for i in 1:dim, j in 1:dim`-loop I mean.

KristofferC:
Ah, no. This also seems to be more expensive for compilation...

src/constructors.jl
```diff
@@ -91,16 +99,8 @@ end
...
-    TensorType = get_base(get_type(S))
-    N = n_components(TensorType)
-    ele = el <: Number ? :(el) : :(el())
-    expr = Expr(:tuple, [ele for i in 1:N]...)
-    quote
-        $(Expr(:meta, :inline))
-        @inbounds return $TensorType($expr)
-    end
-end
+Base.fill{T <: AbstractTensor}(el::Number, S::Type{T}) = apply_all(get_base(T), i -> el)
```
fredrikekre: This became really neat.

src/math_ops.jl
```diff
@@ -58,28 +58,12 @@ julia> det(A)
...
-            $(Expr(:meta, :inline))
-            v = get_data(t)
-            (v[$(idx(1,1))]*(v[$(idx(2,2))]*v[$(idx(3,3))]-v[$(idx(2,3))]*v[$(idx(3,2))]) -
-             v[$(idx(1,2))]*(v[$(idx(2,1))]*v[$(idx(3,3))]-v[$(idx(2,3))]*v[$(idx(3,1))]) +
-             v[$(idx(1,3))]*(v[$(idx(2,1))]*v[$(idx(3,2))]-v[$(idx(2,2))]*v[$(idx(3,1))]))
-        end
-    end
+@inline Base.det(t::SecondOrderTensor{1}) = @inboundsret t[1,1]
+@inline Base.det(t::SecondOrderTensor{2}) = @inboundsret (t[1,1] * t[2,2] - t[1,2] * t[2,1])
+@inline function Base.det(t::SecondOrderTensor{3})
```
fredrikekre: Will this give the same code even for `SymmetricTensor`?

src/tensor_products.jl
```diff
@@ -123,27 +123,13 @@ julia> A ⊗ B
...
+    TensorType(@inline function(i,j,k,l) @inboundsret S1[i,j] * S2[k,l]; end)
```
fredrikekre: Same here; results in the same code for `SymmetricTensor`?

KristofferC: I'm pretty sure  checked that but can check again.

fredrikekre: As long as `compute_index` via `getindex` is elided for `Tensor`, no reason it shouldn't for `SymmetricTensor` as well. Just making sure.

src/math_ops.jl
KristofferC: Appear so, yes.

test/test_ops.jl
```diff
@@ -7,7 +7,7 @@ end
 
 @testsection "tensor ops" begin
 for T in (Float32, Float64, F64), dim in (1,2,3)
-
+println("T = $T, dim = $dim")
```
fredrikekre: Perhaps this should be removed? It was included in #23

src/tensor_products.jl
KristofferC: Yes, exact same code

test/test_ops.jl
KristofferC: It is there because sometimes there were time outs for this section because nothing got printed for 10 minutes. I think maybe when Travis is extra slow. So good to have something that prints maybe?

fredrikekre: But #23 results in prints just as often?

KristofferC: There is no printing there from what I can see?

fredrikekre: Ah, right. It was removed in #23 :) https://github.com/KristofferC/Tensors.jl/pull/23/files#diff-78a3de5eaf1c4bcf42b32436cba105ceL8

KristofferC:
Good to go?

fredrikekre:
I tried this https://github.com/KristofferC/Tensors.jl/pull/19#issuecomment-287619280 for all functions. Test time (eg compilation time) went from 120s to 750. It really helps the compiler to compute the linear index by ourselves. Sadly, cause it led to some nice code deletion

KristofferC:
Perhaps we can make an issue about it...

PR #22, opened 2017-03-17
simd method for dcontract(Sym4, Sym2)

fredrikekre:
```julia
julia> @btime dcontract($AAs, $As);
  8.687 ns (0 allocations: 0 bytes)

julia> @btime dcontract2($AAs, $As);
  4.578 ns (0 allocations: 0 bytes)
```

### DIFF
--- a/src/simd.jl
+++ b/src/simd.jl
@@ -290,6 +290,22 @@ end
         return return Tensor{2, 3}(r)
     end
 end
+@inline function Tensors.dcontract{T <: SIMDTypes}(S1::SymmetricTensor{4, 3, T}, S2::SymmetricTensor{2, 3, T})
+    @inbounds begin
+        D1 = get_data(S1); D2 = get_data(S2)
+        SV11 = tosimd(D1, Val{1},  Val{6})
+        SV12 = tosimd(D1, Val{7}, Val{12})
+        SV13 = tosimd(D1, Val{13}, Val{18})
+        SV14 = tosimd(D1, Val{19}, Val{24})
+        SV15 = tosimd(D1, Val{25}, Val{30})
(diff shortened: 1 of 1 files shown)


KristofferC:
Cool

fredrikekre:
Don't approve things that don't work :p

KristofferC:
The approvement is still conditional on test passes (and we should test this?=

fredrikekre:
Yea, but CI failed. I fixed it now

fredrikekre:
For the record: It was actually a slowdown to do the same for dim = 2.

src/simd.jl
```diff
@@ -290,6 +290,22 @@ end
...
+@inline function Tensors.dcontract{T <: SIMDTypes}(S1::SymmetricTensor{4, 3, T}, S2::SymmetricTensor{2, 3, T})
+    @inbounds begin
+        D1 = get_data(S1); D2 = get_data(S2)
+        SV11 = tosimd(D1, Val{1},  Val{6})
+        SV12 = tosimd(D1, Val{7}, Val{12})
+        SV13 = tosimd(D1, Val{13}, Val{18})
+        SV14 = tosimd(D1, Val{19}, Val{24})
+        SV15 = tosimd(D1, Val{25}, Val{30})
+        SV16 = tosimd(D1, Val{31}, Val{36})
+        D21 = D2[1] * T(1); D22 = D2[2] * T(2); D23 = D2[3] * T(2)
```
fredrikekre: These  ` * T(1)` were pretty retarded. Probably optimized away anyway though...

KristofferC: Yeah, they are there for readability ;)

PR #28, opened 2017-03-21
[WIP] Add to_symmetric_form to reduce order of symmetric tensor

KeitaNakamura:
Hi! Thank you for your awesome package!
I want to add new function to reduce order of symmetric tensor.

```julia
julia> S = rand(SymmetricTensor{2,3})
3×3 Tensors.SymmetricTensor{2,3,Float64,6}:
 0.86066   0.985521  0.444194
 0.985521  0.301275  0.37782
 0.444194  0.37782   0.0806119

julia> to_symmetric_form(S, 1.0) # voigt form for stress tensor
6-element Array{Float64,1}:
 0.86066
 0.301275
 0.0806119
 0.37782
 0.444194
 0.985521

julia> ϵ_voigt = to_symmetric_form(S, 2.0) # voigt form for strain tensor
6-element Array{Float64,1}:
 0.86066
 0.301275
 0.0806119
 0.755639
 0.888387
 1.97104

julia> ϵ_mandel = to_symmetric_form(S, √2) # mandel form for stress and strain tensors
6-element Array{Float64,1}:
 0.86066
 0.301275
 0.0806119
 0.534317
 0.628185
 1.39374

julia> A = rand(SymmetricTensor{4,3});

julia> to_symmetric_form(A ⊡ S, 1.0) ≈ to_symmetric_form(A, 1.0) * ϵ_voigt
true

julia> to_symmetric_form(A ⊡ S, √2) ≈ to_symmetric_form(A, √2) * ϵ_mandel
true
```

If you like this function, I'm going to make its reverse version. :)


KristofferC:
Hello and welcome. Yes, this functionality would be good to have. I started something similar in the old repo, see https://github.com/KristofferC/ContMechTensors.jl/pull/87/files for some inspiration. 

I think good names for these functions are `tovoigt` and `fromvoigt` to go back and forth. It would also be good if this is implemented for non symmetric tensors.

fredrikekre:
Thanks!

I also like `tovoigt` and `fromvoigt` better.
Regarding the scaling of off-diagonal elements, I think it would be nice if it was a bit clearer that the second input argument actually is a scaling. Perhaps with a keyword, defaulting to 1?

```julia
julia> S = rand(SymmetricTensor{2,3});

julia> tovoigt(S) # default scaling = 1
6-element Array{Float64,1}:
 0.86066
 0.301275
 0.0806119
 0.37782
 0.444194
 0.985521

julia> ϵ_voigt = tovoigt(S, offdiagscale=2) # bit clearer what the 2 is actually used for
6-element Array{Float64,1}:
 0.86066
 0.301275
 0.0806119
 0.755639
 0.888387
 1.97104
```

KeitaNakamura:
> I think good names for these functions are tovoigt and fromvoigt to go back and forth.

Okey. I'm going to change it. :)

> It would also be good if this is implemented for non symmetric tensors.

I think this should not be implemented for non symmetric tensors to avoid ambiguity of `fromvoigt` function:

```julia
x = rand(Tensor{2,3})
A = (x + x') / 2 # type is Tensor
S = symmetric(A) # type is SymmetricTensor
voigt_A = tovoigt(A) # type is Array
voigt_S = tovoigt(S) # type is Array
fromvoigt(voigt_A) # type should be Tensor
fromvoigt(voigt_S) # type should be SymmetricTensor
```

KeitaNakamura:
@fredrikekre, I think using the argument as keyword is a good idea.

So the functions that should be implemented are something like this?

```julia
S = rand(SymmetricTensor{2,3})
tovoigt(S) # voigt notation for stress tensor
tovoigt(S, offdiagscale=2) # voigt notation for strain tensor
tovoigt(S, offdiagscale=√2) # mandel notation for stress and strain tensors (this is not voigt notation, actually)

scale = 2
fromvoigt(tovoigt(S)) ≈ S # true
fromvoigt(tovoigt(S, offdiagscale=scale), offdiagscale=1/scale) ≈ S # true
```

fredrikekre:
> So the functions that should be implemented are something like this?

Yes, looks good to me.

KristofferC:
> I think this should not be implemented for non symmetric tensors to avoid ambiguity of `fromvoigt` function:

This can easily be fixed by giving a type parameter as the first argument, like `fromvoigt(Tensor{2, 3}, voigt_A)`. In order for the function `fromvoigt` to be type stable, you will need to have a type parameter anyway because the dimension of the resulting tensor can not be inferred by the type of the array.

KristofferC:
Also, I wonder if the `offdiagscale` should be implemented so `fromvoigt(tovoigt(S, offdiagscale=scale), offdiagscale=scale) ≈ S`. The `scale` parameter then tells what the scaling is on the off diagonal **in Voigt form** and is thus the same whether you use `fromvoigt` or `tovoigt`.

KeitaNakamura:
I modified `tovoigt` function as you suggested. Also, it can be used for non symmetric tensors.

```julia
julia> A = rand(Tensor{4,3});

julia> x = rand(Tensor{2,3});

julia> tovoigt(A) * tovoigt(x) ≈ tovoigt(A ⊡ x)
true

julia> A_sym = symmetric(A);

julia> x_sym = symmetric(x);

julia> tovoigt(A_sym) * tovoigt(x_sym, offdiagscale=2) ≈ tovoigt(A_sym ⊡ x_sym)
true
```

KeitaNakamura:
The difference in code between `Tensor` and `SymmetricTensor` would be only range for `for` loop. Should I make it simple by using macro? For example:

```julia
for (TensorType, j0) in ((:(Tensor{2,dim,T,M}), 1),
                         (:(SymmetricTensor{2,dim,T,M}), :i))
    @eval function tovoigt{dim,T,M}(A::$TensorType; offdiagscale::Real=1)
        v = zeros(T, M)
        for i in 1:dim, j in $j0:dim
            idx, scale = _index_scale(dim, i, j, offdiagscale)
            v[idx] = scale * A[i,j]
        end
        return v
    end
end

for (TensorType, j0, l0) in ((:(Tensor{4,dim,T,M}), 1, 1),
                             (:(SymmetricTensor{4,dim,T,M}), :i, :k))
    @eval function tovoigt{dim,T,M}(A::$TensorType; offdiagscale::Real=1)
        n = Int(√M)
        v = zeros(T, n, n)
        for i in 1:dim, j in $j0:dim, k in 1:dim, l in $l0:dim
            I, si = _index_scale(dim, i, j, offdiagscale)
            J, sj = _index_scale(dim, k, l, offdiagscale)
            v[I,J] = si * sj * A[i,j,k,l]
        end
        return v
    end
end
```

KristofferC:
You can do that or use a helper function like:

```
dimrange{order, dim}(::Tensor{order, dim}, i) = dim
dimrange(::SymmetricTensor, i) = i
```

and then have this in your loop:

```
for i in 1:dim, j in 1:dimrange(A, i), k in 1:dim, l in 1:dimrange(A, k)
```

fredrikekre:
Also, `tovoigt`/`fromvoigt` for `Tensor` should not need the scaling factor? So thats different as well.

fredrikekre:
BTW, it would  be nice if you could add some tests as well. The examples in this conversation would serve as good test-cases!

KeitaNakamura:
@KristofferC, Oh, yes, sorry for my stupid question.

@fredrikekre, I would remove scaling factor from functions for `Tensor`.

This is my first contribution, so if I do something wrong, please let me know. :)

test/test_ops.jl
```diff
@@ -251,5 +251,20 @@ end # of testsection
     @test rotate(a, b, π) ≈ rotate(a, b, -π)
     @test rotate(a, b, π/2) ≈ rotate(a, -b, -π/2)
 end
+
+@testsection "tovoigt/fromvoigt" begin
```
KristofferC: Could you try this for more than dimension `3` i.e. by adding a loop `for dim in 1:3` and have `A = rand(Tensor{4,dim})` etc?

KristofferC:
Thanks for the added test! I pushed a commit to this branch with a bit of added documentation and "compacting" a few of the functions. Hope you don't mind.

KristofferC:
AV failure on 32 bit. Should probably disable strict doctests on 32 bit systems... Let's deal with it later.

KristofferC:
Thanks for the contribution!

KeitaNakamura:
My pleasure. :)

PR #32, opened 2017-03-23
Add inv for fourth order tensors

KeitaNakamura:
This PR is to support `inv` for fourth order tensors. This is useful in some computations of continuum mechanics. The definitions of inverse of fourth order tensors are follows:

```julia
julia> A = rand(Tensor{4,3});

julia> A ⊡ inv(A) ≈ one(Tensor{4,3})
true

julia> S = rand(SymmetricTensor{4,3});

julia> S ⊡ inv(S) ≈ one(SymmetricTensor{4,3})
true
```

### DIFF
--- a/src/math_ops.jl
+++ b/src/math_ops.jl
@@ -153,6 +153,16 @@ end
     end
 end
 
+function Base.inv{dim}(t::Tensor{4, dim})
+    fromvoigt(Tensor{4, dim}, inv(tovoigt(t)))
+end
+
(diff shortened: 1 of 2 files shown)


KristofferC:
Hmm, maybe more examples for the documentation ;) ;)

KristofferC:
Thanks again! This was very much needed.

KeitaNakamura:
My pleasure 😀

PR #33, opened 2017-03-23
actually build docs

KristofferC:
See https://travis-ci.org/KristofferC/Tensors.jl/jobs/214207279#L418


KristofferC:
Not since https://github.com/KristofferC/Tensors.jl/commit/49d600a05c07407f410bdf74ebe9eae444748929 I think

PR #34, opened 2017-03-23
Fix mean and dev

KeitaNakamura:
Currently `dev` function doesn't return a correct result for 1 and 2 dimension tensors because `3` is always used to compute volumetric part instead of `dim`.

```julia
julia> rand(Tensor{2,2}) |> dev |> trace
0.07594753985159675

julia> rand(Tensor{2,1}) |> dev |> trace
0.6577749573412563
```

This PR:
```julia
julia> rand(Tensor{2,2}) |> dev |> trace
0.0

julia> rand(Tensor{2,1}) |> dev |> trace
0.0
```

I also fixed `mean` function to use `dim` instead of `3`.

NOTE: I wanted to add tests to check this, but it sometimes failed:

```julia
julia> rand(Tensor{2,3}) |> dev |> trace ≈ 0.0
false
```

### DIFF
--- a/src/math_ops.jl
+++ b/src/math_ops.jl
@@ -298,7 +298,7 @@ julia> trace(A)
 end
 vol(S::SecondOrderTensor) = trace(S)
 
-Base.mean(S::SecondOrderTensor) = trace(S) / 3
+Base.mean{dim}(S::SecondOrderTensor{dim}) = trace(S) / dim
 
 """
 ```julia
@@ -323,7 +323,7 @@ julia> trace(dev(A))
 """
 @inline function dev(S::SecondOrderTensor)
     Tt = get_base(typeof(S))
-    tr = trace(S) / 3
+    tr = mean(S)
     Tt(
         @inline function(i, j)
             @inbounds  v = i == j ? S[i,j] - tr : S[i,j]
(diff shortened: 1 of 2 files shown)


KristofferC:
We have had this before. It was changed at: https://github.com/KristofferC/Tensors.jl/commit/a62fa904c51b175e3039bbd5f3f06890e0f2c66d.

Do you have a reference for your definition of `dev`? AFAIU it is typically defined as `a_ii / 3` no matter the dimension.

KeitaNakamura:
I think that the most important aspect of this additive decomposition is that the volumetric part is orthogonal to the deviatoric part, i.e, their double contraction should be 0. But, I also couldn't find a reference. I will try to find it. :)

KeitaNakamura:
I couldn't find a reference. Basically we have to choose plain stress or plain stress for 2D analysis because 2D is a special case of 3D, so it's hard to discuss with this issue. I will close this PR as I may misunderstand.

fredrikekre:
I mean, things are so weird in 2D. Arguably `dev` should only be defined for 3D tensors? I only found a reference for 3D, which states the identities in the OP, that the trace of a deviatoric tensor should be 0.

Regarding `mean`, if we have plain stress, that would mean the third normal stress is 0, and then `mean(A) = (A11 + A22 + 0)/3`, i.e. as we have it now. For plain strain, the stress tensor is not 2D and then the current behaviour would be ok.

fredrikekre:
What I found was this definition, as @KristofferC  mentioned, `a_ij - δ_ij a_kk / 3` but I *think* they are only talking about 3D tensors.

KeitaNakamura:
I think that removing `dev` from tensors with 1 and 2 dimensions is better. What `vol(A) ⊡ dev(A)` is not zero makes me confused (where `vol` is defined in #35).

PR #35, opened 2017-03-23
Modify vol to return tensor

KeitaNakamura:
Currently `vol` function is the same as `trace`, but it is more useful to return tensor. This is consistent with `dev` function.

```julia
julia> A = rand(Tensor{2,3})
3×3 Tensors.Tensor{2,3,Float64,9}:
 0.349248  0.383259    0.493558
 0.327233  0.0432712   0.0638624
 0.687384  0.00967807  0.654179

julia> vol(A) + dev(A) ≈ A
true
```

### DIFF
--- a/docs/src/man/other_operators.md
+++ b/docs/src/man/other_operators.md
@@ -112,7 +112,7 @@ The skew symmetric part of a symmetric tensor is zero.
 skew
 ```
 
-## Deviator
+## Deviatoric tensor
 
 The deviatoric part of a second order tensor is defined by
 
@@ -120,6 +120,16 @@ $\mathbf{A}^\text{dev} = \mathbf{A} - \frac{1}{3} \mathrm{trace}[\mathbf{A}] \ma
 
 ```@docs
 dev
+```
+
+## Volumetric tensor
+
+The volumetric part of a second order tensor is defined by
+
+$\mathbf{A}^\text{vol} = \frac{1}{3} \mathrm{trace}[\mathbf{A}] \mathbf{I} \Leftrightarrow A_{ij}^\text{vol} = \frac{1}{3}A_{kk}\delta_{ij}$
+
+```@docs
+vol
 ```
 
 ## Cross product
--- a/src/math_ops.jl
+++ b/src/math_ops.jl
@@ -275,7 +275,6 @@ Extract eigenvectors from an `Eigen` object, returned by [`eigfact`](@ref).
 trace(::SecondOrderTensor)
 ```
 Computes the trace of a second order tensor.
-The synonym `vol` can also be used.
 
 **Example:**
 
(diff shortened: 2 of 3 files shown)


KristofferC:
Do you have a reference for this? For example Wikipedia: https://en.wikipedia.org/wiki/Infinitesimal_strain_theory#Volumetric_strain and http://www.eduresourcecollection.com/civil_sm_Strains.php indicates that the volumetric strain is a scalar (it is the ratio of two volumes).

KeitaNakamura:
This is just for convenience. The `vol` in PR means "volumetric part" of tensor, which is consistent with `dev` because `dev` returns "deviatoric part" of tensor. If we think about stress, `vol(stress)` returns hydrostatic stress tensor.

KristofferC:
Ok, that makes sense to me. I'm ok with this. @fredrikekre?

fredrikekre:
That wikipedia article also mentions "volumetric strain tensor" though, as defined in this PR. This https://biomechanics.stanford.edu/me337/me337_s03.pdf also defines like this.
I think vol might be better than this, than as a synonym to trace.

KristofferC:
I rather not restrict this to 3d right now.. We are clear with what the definition is in the documentation.

fredrikekre:
Okay. They are effectively useless in 1 and 2 dim though.

```julia
julia> A2 = rand(SymmetricTensor{2, 2})
2×2 Tensors.SymmetricTensor{2,2,Float64,3}:
 0.491819  0.868985
 0.868985  0.798158

julia> A3 = SymmetricTensor{2, 3}((i,j) -> i<3 && j<3 ? A2[i, j] : 0.0)
3×3 Tensors.SymmetricTensor{2,3,Float64,6}:
 0.491819  0.868985  0.0
 0.868985  0.798158  0.0
 0.0       0.0       0.0

julia> mise2d = sqrt(3/2 * dev(A2) ⊡ dev(A2))
1.5730414362349892

julia> mise3d = sqrt(3/2 * dev(A3) ⊡ dev(A3))
1.658854850176772
```

KristofferC:
Ok, I agree we can remove them from 3d but then we should have an easy way to go from 2d to 3d and back again so someone can do `2d(dev(3d(A2)))`

fredrikekre:
We could make `dev` always return a 3d tensor?
But when going back to 2d, you think we should just cut off A33? That is a bit weird.

KristofferC:
Hmm, you would just do `norm(dev(3d))` so no need to go back to 2e.

fredrikekre:
Sure. I would make `dev(Tensor{2, 2})` -> `Tensor{2, 3}` in that case so the conversion to 3d is implicit. We do this for `cross` already (since `cross` is also weird in 1/2d):
```julia
julia> v = rand(Vec{2})
2-element Tensors.Tensor{1,2,Float64,2}:
 0.0349476
 0.383921 

julia> v × v
3-element Tensors.Tensor{1,3,Float64,3}:
 0.0
 0.0
 0.0
```

KristofferC:
That makes sense to me.

PR #39, opened 2017-03-25
make AD work for scalars

KristofferC:


src/automatic_differentiation.jl
```diff
@@ -308,21 +322,21 @@ julia> ∇f = gradient(norm, A)
 julia> ∇f, f = gradient(norm, A, :all);
 ```
 """
-function Base.gradient{F}(f::F, v::Union{SecondOrderTensor, Vec})
+function Base.gradient{F <: Function}(f::F, v::Union{SecondOrderTensor, Vec, Real})
```
KristofferC: Har to do this `<: Function` because of an ambiguity with some function in `Calculus` otherwise... We be type piracying all over the place..

KeitaNakamura:
I think this still doesn't work for scalar because input type must be the same as output type in `_extract_gradient` (as we discussed with #40).

Maybe we should make `derivative` function like ForwardDiff package? For example

```julia
@inline function _extract_derivative{N, T}(v::Dual{N, T}, ::Real)
    return partials(v, 1)
end

@generated function _extract_derivative{order, dim, D <: Dual}(v::AbstractTensor{order, dim, D}, ::Real)
    TensorType = get_base(v)
    nc = n_components(TensorType)
    exps = Expr(:tuple, [:(partials(v[$i], 1)) for i in 1:nc]...)
    return quote
        $(Expr(:meta, :inline))
        @inbounds return $TensorType($exps)
    end
end
```

KristofferC:
To clarify, this has nothing to do with Tensors at all, it is purely for scalars. I just thought it could perhaps be a good functionality to provide.

KeitaNakamura:
Sorry, I just noticed that. This functionality is really nice. How do you think to implement AD for scalar->tensor function.

KristofferC:
It is not hard:

```
julia> f(x) = Vec{3}((x, x^2, sin(x)))                                                                                                                                             
f (generic function with 1 method)                                                                                                                                                 
                                                                                                                                                                                   
julia> dualx = Tensors._load(2.0)                                                                                                                                                  
Dual(2.0,1.0)                                                                                                                                                                      
                                                                                                                                                                                   
julia> f(dx)                                                                                                                                                                       
3-element Tensors.Tensor{1,3,ForwardDiff.Dual{1,Float64},3}:                                                                                                                       
 Dual(2.0,1.0)                                                                                                                                                                     
 Dual(4.0,4.0)                                                                                                                                                                     
 Dual(0.909297,-0.416147) 
```

we just need a function that extracts the partials from that result (by knowing that the input argument was a scalar).

KeitaNakamura:
I see. So changing `_extract_derivative` to `_extract_gradient` in my comment may work?

KeitaNakamura:
I checked it. If I add following code,

```julia
for TensorType in (Tensor, SymmetricTensor)
    @eval @generated function _extract_gradient{order, dim, D <: Dual, N, T2 <: Real}(v::$TensorType{order, dim, D, N}, ::T2)
        RetType = $TensorType{order,dim}
        exps = Expr(:tuple, [:(partials(v[$i], 1)) for i in 1:N]...)
        return quote
            $(Expr(:meta, :inline))
            @inbounds return $RetType($exps)
        end
    end
end
```

I can use AD for scalar->tensor function like

```julia
julia> f = x -> Tensor{2,2}((i,j) -> x^(i+j))
(::#1) (generic function with 1 method)

julia> gradient(f, 2)
2×2 Tensors.Tensor{2,2,Int64,4}:
  4  12
 12  32
```

KristofferC:
That's what I am thinking too... It was just so little extra code to support and it seems like it might pop up in applications where you would otherwise typically use this package.

KeitaNakamura:
In my research so far, I have never used scalar -> tensor function. 🙁
But someone may try this? For example:

```julia
function f(x::Tensor, y::Real)
    ...
    return ret # ret is Tensor
end

gradient(x -> f(x, y), rand(Tensor{2,3})) # ∂f∂x
gradient(y -> f(x, y), 3.0) # ∂f∂y
```

KeitaNakamura:
Could we merge this PR? I think that we could ignore scalar -> tensor case for now.

KristofferC:
I'm still not sure we should do this. It's basically just replicating functionality from ForwardDiff (`ForwardDiff.derivative(f, x)`).

KeitaNakamura:
> It's basically just replicating functionality from ForwardDiff (ForwardDiff.derivative(f, x)).

Yes, but it would be useful for me if I could just use `Tensors.jl`'s functions all the way. In this case, I suppose that the functionality of Tensors.jl is more elegant than ForwardDiff because we could always use `gradient` for first-order differentiation. Is there any disadvantage?

KristofferC:
The disadvantage is that we are "type piracing", which is that we extend a Base method with arguments not defined in the package.

KeitaNakamura:
Okey. How about changing `gradient` to `grad` or something like that?

KeitaNakamura:
Hmm... It also seems not good actually.

KristofferC:
We could just have a `gradient` that is not extending base. But then you need to write `Tensors.gradient` etc.

fredrikekre:
You can do this in your own code if you want the same API:

```julia
julia> using Tensors

julia> @eval Tensors begin
           import ForwardDiff
           Base.gradient{F<:Function}(f::F, x::Real) = ForwardDiff.derivative(f, x)
           Base.gradient{F<:Function}(f::F, x::Real, ::Symbol) = (ForwardDiff.derivative(f, x), f(x))
       end

julia> gradient(sin, 0.5)
0.8775825618903728

julia> gradient(sin, 0.5, :all)
(0.8775825618903728, 0.479425538604203)
```

KeitaNakamura:
@KristofferC, Yeah, it's the same way as ForwardDiff etc, but I also don't want to add prefix `Tensors.` to `gradient` function every time. That's why I suggested using different name instead of `gradient`.

@fredrikekre, Thank you for your suggestion. I already did the similar way, but your code is much simpler than mine. I would follow this way for now.

Thank you again, both of you. :)

Issue #26, opened 2017-03-20
SIMD promotion


fredrikekre:
It is possible to play with promotion a bit (for tensor products in particular) to streamline more calculations into `simd.jl`. This could potentially speed up things, if the promotion is cheap enough.

fredrikekre:
closed by #51

PR #41, opened 2017-03-25
remove unnecessary promotions

KristofferC:
Let things promote at the elementwise level instead.

Benchmark:

```jl
using Tensors

function u_ana{T}(x::Vec{2, T}) 
    xs = (Vec{2}((-0.5,  0.5)),
          Vec{2}((-0.5, -0.5)),
          Vec{2}(( 0.5,  -0.5)))
    σ = 1/8
    s = zero(eltype(x))
    for i in 1:3
        s += exp(- norm(x - xs[i])^2 / σ^2)
    end
    return max(1e-15 * one(T), s) # Denormals, be gone
end

const ud2 = Tensors._load(rand(Vec{2}))
```

## Master 0.5:
```
julia> @btime u_ana($ud2)
  422.688 ns (12 allocations: 912 bytes)
```
## Master 0.6:
```
julia> @btime u_ana($ud2)
  87.265 ns (0 allocations: 0 bytes)
```
## PR:
```
julia> @btime u_ana($ud2)
  71.876 ns (0 allocations: 0 bytes)
```

### DIFF
--- a/src/basic_operations.jl
+++ b/src/basic_operations.jl
@@ -4,30 +4,31 @@
 
 # Unary
 @inline Base.:+{T <: AbstractTensor}(S::T) = S
-@inline Base.:-{T <: AbstractTensor}(S::T) = map(-, S)
+@inline Base.:-{T <: AbstractTensor}(S::T) = _map(-, S)
 
 # Binary
-@inline Base.:+{order, dim, T}(S1::Tensor{order, dim, T}, S2::Tensor{order, dim, T}) = map(+, S1, S2)
-@inline Base.:+{order, dim, T}(S1::SymmetricTensor{order, dim, T}, S2::SymmetricTensor{order, dim, T}) = map(+, S1, S2)
-@inline Base.:+{order, dim}(S1::AbstractTensor{order, dim}, S2::AbstractTensor{order, dim}) = +(promote(S1, S2)...)
-@inline Base.:-{order, dim, T}(S1::Tensor{order, dim, T}, S2::Tensor{order, dim, T}) = map(-, S1, S2)
-@inline Base.:-{order, dim, T}(S1::SymmetricTensor{order, dim, T}, S2::SymmetricTensor{order, dim, T}) = map(-, S1, S2)
-@inline Base.:-{order, dim}(S1::AbstractTensor{order, dim}, S2::AbstractTensor{order, dim}) = -(promote(S1, S2)...)
+@inline Base.:+{order, dim}(S1::Tensor{order, dim}, S2::Tensor{order, dim}) = _map(+, S1, S2)
+@inline Base.:+{order, dim}(S1::SymmetricTensor{order, dim}, S2::SymmetricTensor{order, dim}) = _map(+, S1, S2)
+@inline Base.:-{order, dim}(S1::Tensor{order, dim}, S2::Tensor{order, dim}) = _map(-, S1, S2)
+@inline Base.:-{order, dim}(S1::SymmetricTensor{order, dim}, S2::SymmetricTensor{order, dim}) = _map(-, S1, S2)
+@inline Base.:-{order, dim}(S1::AnyTensor{order, dim}, S2::AnyTensor{order, dim}) =  ((SS1, SS2) = promote(S1, S2); _map(-, SS1, SS2))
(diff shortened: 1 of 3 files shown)


src/basic_operations.jl
```diff
@@ -7,27 +7,28 @@
...
 @inline function Base.map{T <: AbstractTensor}(f, S::T)
     return apply_all(S, @inline function(i) @inboundsret f(S.data[i]); end)
 end
 
-@inline Base.map{order, dim, T1, T2}(f, S1::AbstractTensor{order, dim, T1}, S2::AbstractTensor{order, dim, T2}) = ((SS1, SS2) = promote(S1, S2); map(f, SS1, SS2))
+Base.map{order, dim}(f, S1::Tensor{order, dim}, S2::SymmetricTensor{order, dim}) = ((SS1, SS2) = promote(S1, S2); map(f, SS1, SS2))
+Base.map{order, dim}(f, S1::SymmetricTensor{order, dim}, S2::Tensor{order, dim}) = ((SS1, SS2) = promote(S1, S2); map(f, SS1, SS2))
 
-@inline function Base.map{T <: AllTensors}(f, S1::T, S2::T)
+@inline function Base.map(f, S1::AllTensors, S2::AllTensors)
```
fredrikekre: Should this not be restricted to same order and dim? On this PR this works:

```julia
julia> A = rand(Tensor{2, 2}); B = rand(Tensor{2, 3});

julia> map(+, A, B)
2×2 Tensors.Tensor{2,2,Float64,4}:
 1.31342  1.02726 
 1.28085  0.807102
```

fredrikekre:
We had this promotion path before #5 actually. This is the other way around of #26 , but we can add separate promotion for `T <: SIMDTypes` so this is fine.

src/basic_operations.jl
KristofferC: Hm, yeah. I sort of saw this as an implementation detail (that you would never call `map` yourself). Perhaps we call it `_map` or something? Or do we want to support the official `map`.

KristofferC:
To be explicit about the reason for this PR. Promoting when only the element types are different is unnecessary because for example, promoting `Float64 * Dual` to `Dual * Dual` means that all operations are more expensive.

fredrikekre:
Yea that makes sense, but should restrict `order` and `dim` in `map`

src/basic_operations.jl
fredrikekre: `_map` I guess.

src/basic_operations.jl
```diff
@@ -4,30 +4,31 @@
...
+Base.:-(S1::AbstractTensor, S2::AbstractTensor) = throw(DimensionMismatch("dimension and order must match"))
 
 # map implementations
-@inline function Base.map{T <: AbstractTensor}(f, S::T)
+@inline function _map{T <: AbstractTensor}(f, S::T)
     return apply_all(S, @inline function(i) @inboundsret f(S.data[i]); end)
 end
 
-@inline Base.map{order, dim, T1, T2}(f, S1::AbstractTensor{order, dim, T1}, S2::AbstractTensor{order, dim, T2}) = ((SS1, SS2) = promote(S1, S2); map(f, SS1, SS2))
+_map{order, dim}(f, S1::Tensor{order, dim}, S2::SymmetricTensor{order, dim}) = ((SS1, SS2) = promote(S1, S2); _map(f, SS1, SS2))
```
fredrikekre: This and the next line is not used anymore (see test-coverage) since they are replaced by 14-15 with `AnyTensor`.
Related: Should we have promotion methods that promote `SymmetricTensor` -> `Tensor` but does not promote eltype?

fredrikekre:
What does `AnyTensor` do that `AbstractTensor` does not?

fredrikekre:
Note that promotion on element level is already done for all the tensor products, this was only done for +/-

fredrikekre:
PR:
```julia
julia> @btime u_ana($ud2)
  62.931 ns (0 allocations: 0 bytes)
```
Master:
```julia
julia> @btime u_ana($ud2)
  76.403 ns (0 allocations: 0 bytes)
```

fredrikekre:
Hmmm... This actually slowed down some cases which would be promoted to SIMDTensor before. So #26 should be fixed.

PR #42, opened 2017-03-26
Fix #40

### LINKED ISSUE #40: `symmetric` breaks automatic differentiation
```julia
julia> Isym = one(SymmetricTensor{4,3});

julia> gradient(symmetric, rand(Tensor{2,3})) ≈ Isym
false

julia> gradient(symmetric, rand(SymmetricTensor{2,3})) ≈ Isym
true

julia> gradient(x -> (x+x')/2, rand(Tensor{2,3})) ≈ Isym
true

julia> gradient(x -> (x+x')/2, rand(SymmetricTensor{2,3})) ≈ Isym
true
```

This is because `symmetric` changes type of argument from `Tensor` to `SymmetricTensor`, and AD in Tensors doesn't support this change in `_extract_gradient`. I'm not sure that this is really problem in this package. Should I carefully use by myself?

KeitaNakamura:


KeitaNakamura:
How should I fix the ambiguity of function?

src/automatic_differentiation.jl
```diff
@@ -195,6 +195,25 @@ end
     return ∇f
 end
 
+# for non dual variable
+@inline function _extract_value(v::Any, ::Any)
+    return v
+end
+for TensorType in (Tensor, SymmetricTensor)
```
fredrikekre: Perhaps this is simpler just as:

```julia
@inline _extract_gradient{order, dim, T}(v::T, ::Tensor{order, dim}) = zero(Tensor{order, dim, T})
@inline _extract_gradient{order, dim, T}(v::T, ::SymmetricTensor{order, dim}) = zero(SymmetricTensor{order, dim, T})
@inline _extract_gradient{order, dim, T}(v::Tensor{order, dim, T}, ::Tensor{order, dim}) = zero(Tensor{2order, dim, T})
@inline _extract_gradient{order, dim, T}(v::SymmetricTensor{order, dim, T}, ::SymmetricTensor{order, dim}) = zero(SymmetricTensor{2order, dim, T})
```

KristofferC: Will `2order` be type stable?

fredrikekre: No...

fredrikekre: Nevermind then

PR #43, opened 2017-03-26
Add tomandel and frommandel

KeitaNakamura:
I often use "Mandel" form to reduce the order of symmetric tensors because I don't need to care that the tensor is stress-like tensor or strain-like tensor. I think "Mandel" form is as popular as "Voigt" form, so it would be nice if this package has `tomandel` and `frommandel`.

### DIFF
--- a/src/Tensors.jl
+++ b/src/Tensors.jl
@@ -13,7 +13,7 @@ export tdot, dott, dotdot
 export hessian#, gradient
 export basevec, eᵢ
 export rotate
-export tovoigt, fromvoigt
+export tovoigt, fromvoigt, tomandel, frommandel
 
 #########
 # Types #
--- a/src/math_ops.jl
+++ b/src/math_ops.jl
@@ -158,9 +158,7 @@ function Base.inv{dim}(t::Tensor{4, dim})
 end
 
 function Base.inv{dim, T}(t::SymmetricTensor{4, dim, T})
-    # use mandel form to reduce order of symmetric tensor
-    s = T(√2)
-    fromvoigt(SymmetricTensor{4, dim}, inv(tovoigt(t, offdiagscale=s)), offdiagscale=s)
+    frommandel(SymmetricTensor{4, dim}, inv(tomandel(t)))
 end
 
 """
--- a/src/voigt.jl
+++ b/src/voigt.jl
@@ -7,7 +7,8 @@ dimrange(::SymmetricTensor, i, dim) = i
 Converts a tensor to "Voigt"-format using the following index order:
 `[11, 22, 33, 23, 13, 12, 32, 31, 21]`.
 For `SymmetricTensor`s, the keyword argument `offdiagscale` sets a scaling factor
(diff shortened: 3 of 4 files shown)


KeitaNakamura:
This is a reference of "Mandel" form:
[https://en.wikipedia.org/wiki/Voigt_notation](https://en.wikipedia.org/wiki/Voigt_notation)

KristofferC:
Out of curiosity, why do you need to use Voigt / Mandel form at all instead of just using `Tensor`s all the way?

KeitaNakamura:
Some computations force me to use Voigt / Mandel form, for example, the return-mapping algorithm in this book ([Computational Inelasticity](http://www.springer.com/us/book/9780387975207), see BOX 3.5 (General Closest Point Projection Iteration) in pp.146). To solve simultaneous linear equations including fourth order tensors, we first need to reduce the order of tensor to get matrix form.

fredrikekre:
I was also curious about that!
But that’s basically what you did this friday @KristofferC :p

On a separate note: if you use `tovoigt` and friends in an iterative algorithm like that they should probably be implemented with in-place versions? `tovoigt!` etc

KeitaNakamura:
Something like this?

```julia
julia> A = zeros(12,12);

julia> Isym = one(SymmetricTensor{4,3});

julia> tomandel!(A, 1:6, 1:6, Isym)
12×12 Array{Float64,2}:
 1.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0
 0.0  1.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0
 0.0  0.0  1.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0
 0.0  0.0  0.0  1.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0
 0.0  0.0  0.0  0.0  1.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0
 0.0  0.0  0.0  0.0  0.0  1.0  0.0  0.0  0.0  0.0  0.0  0.0
 0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0
 0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0
 0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0
 0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0
 0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0
 0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0
```

I think that if we can also easily extract original tensor `Isym` from `A`, it would be very nice. :)

PR #46, opened 2017-03-28
Add feature: differentiation by prime notation

KeitaNakamura:
This PR is to support the prime notation for differentiation. In Tensors.jl package, we can always use `gradient` for differentiation, which means if we have the prime notation like [Calculus.jl](https://github.com/johnmyleswhite/Calculus.jl), we can use it for all of functions. Although this feature is not necessary for computations, we can improve readability of code.

```julia
julia> A = rand(SymmetricTensor{2, 2});

julia> norm'(A) ≈ gradient(norm, A)
true

julia> norm''(A) ≈ hessian(norm, A)
true

julia> E = rand(Tensor{4,2});

julia> ψ(ϵ) = 1/2 * ϵ ⊡ E ⊡ ϵ;

julia> ϵ = rand(Tensor{2,2});

julia> ψ''(ϵ)  ≈ hessian(ψ, ϵ)
true
```

However, someone doesn't like this feature and may have their own `ctranspose` definition for functions. So I added a macro `@use_prime_diff` to enable this feature.

```julia
julia> using Tensors

julia> @which trace'
ctranspose(f::Function) at /Users/nakamura/.julia/v0.5/Calculus/src/derivative.jl:26
```

```julia
julia> using Tensors

julia> Tensors.@use_prime_diff

julia> @which trace'
ctranspose{#84#F<:Function}(f::#84#F) at /Users/nakamura/.julia/v0.5/Tensors/src/prime_differentiation.jl:27
```

Also I added one more feature for functions having several arguments.

```julia
julia> f(x, y) = trace(x) + norm(y)
f (generic function with 1 method)

julia> A = rand(Tensor{2,3}); B = rand(Tensor{2,3});

julia> f'(By{1}, A, B) # differentiate `f` by `x` (first argument).
3×3 Tensors.Tensor{2,3,Float64,9}:
 1.0  0.0  0.0
 0.0  1.0  0.0
 0.0  0.0  1.0

julia> f'(By{2}, A, B); # differentiate `f` by `y` (second argument).

jjulia> gradient(x -> gradient(y -> f(A, y), x), B) ≈ (y -> f(A, y))''(B) ≈ f''(By{2}, A, B)
true
```

### DIFF
--- a/src/prime_differentiation.jl
+++ b/src/prime_differentiation.jl
@@ -23,8 +23,8 @@ end
 
 @inline (p::PrimeDiff)(x) = gradient(p.f, x)
 
-macro use_prime_diff()
-    :(@inline Base.ctranspose{F <: Function}(f::F) = PrimeDiff{1}(f))
+function use_prime_diff()
+    @inline Base.ctranspose{F <: Function}(f::F) = Tensors.PrimeDiff{1}(f)
 end
 
 immutable By{N} end
--- a/src/prime_differentiation.jl
+++ b/src/prime_differentiation.jl
@@ -1,0 +1,50 @@
+immutable PrimeDiff{N, F}
+    f::F
+end
+
+@generated function (::Type{PrimeDiff{N}}){N, F}(f::F)
+    ex = :(PrimeDiff{1, F}(f))
+    for i in 2:N
+        ex = :(_add_prime($ex))
+    end
+    return quote
+        $(Expr(:meta, :inline))
+        $ex
+    end
+end
+
+@generated function _add_prime{N}(p::PrimeDiff{N})
+    return quote
+        $(Expr(:meta, :inline))
+        PrimeDiff{$(N+1), typeof(p)}(p)
+    end
+end
+@inline Base.ctranspose{N}(p::PrimeDiff{N}) = _add_prime(p)
+
+@inline (p::PrimeDiff)(x) = gradient(p.f, x)
+
+macro use_prime_diff()
+    :(@inline Base.ctranspose{F <: Function}(f::F) = PrimeDiff{1}(f))
+end
+
+immutable By{N} end
+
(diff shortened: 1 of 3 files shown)


KeitaNakamura:
In v0.6, we cannot create a new function in `@generated` function ([https://github.com/JuliaLang/julia/issues/19299](https://github.com/JuliaLang/julia/issues/19299))? In that case, `By{N}` feature should be removed.

KeitaNakamura:
I realized that `use_prime_diff` doesn't need to be a macro.

```julia
julia> using Tensors

julia> Tensors.use_prime_diff()

julia> @which trace'
ctranspose{#84#F<:Function}(f::#84#F) at /Users/nakamura/.julia/v0.5/Tensors/src/prime_differentiation.jl:27
```

src/prime_differentiation.jl
```diff
@@ -0,0 +1,50 @@
...
+        $(Expr(:meta, :inline))
+        PrimeDiff{$(N+1), typeof(p)}(p)
+    end
+end
+@inline Base.ctranspose{N}(p::PrimeDiff{N}) = _add_prime(p)
+
+@inline (p::PrimeDiff)(x) = gradient(p.f, x)
+
+function use_prime_diff()
+    @inline Base.ctranspose{F <: Function}(f::F) = Tensors.PrimeDiff{1}(f)
```
fredrikekre: This will just overwrite the method from Calculus, so after this the Calculus version is gone. This is also type piracy (function piracy?) and I think it is very strange that Calculus.jl are allowed to define this in the first place. So I'm not sure about this...

KeitaNakamura:
> prime notation are typically only used for derivative scalar->scalar

I definitely agreed with your opinion. Also the implementation seems to be not good. I would close this PR. Thank you. :)

PR #53, opened 2017-04-07
Add an integer power function for second order tensors

KeitaNakamura:
This would be useful in some cases.

```julia
julia> t = rand(Tensor{2,3});

julia> t^0 ≈ one(t)
true

julia> t^1 ≈ t
true

julia> t^2 ≈ t ⋅ t
true

julia> t^3 ≈ t ⋅ t ⋅ t
true

julia> t^-1 ≈ inv(t)
true
```

### DIFF
--- a/src/basic_operations.jl
+++ b/src/basic_operations.jl
@@ -18,6 +18,22 @@
 @inline Base.:*(S::AbstractTensor, n::Number) = _map(x -> x*n, S)
 @inline Base.:*(n::Number, S::AbstractTensor) = _map(x -> n*x, S)
 @inline Base.:/(S::AbstractTensor, n::Number) = _map(x -> x/n, S)
+function Base.:^(S::SecondOrderTensor, p::Int)
+    if p == 1
+        return S
+    elseif p == 0
+        return one(S)
+    elseif p == -1
+        return inv(S)
+    elseif p < 0
+        throw(DomainError())
+    end
+    t = S
+    for _ in 2:p
+        t = t ⋅ S
+    end
+    return t
+end
 
 Base.:+(S1::AbstractTensor, S2::AbstractTensor) = throw(DimensionMismatch("dimension and order must match"))
 Base.:-(S1::AbstractTensor, S2::AbstractTensor) = throw(DimensionMismatch("dimension and order must match"))
--- a/src/basic_operations.jl
+++ b/src/basic_operations.jl
@@ -18,15 +18,15 @@
 @inline Base.:*(S::AbstractTensor, n::Number) = _map(x -> x*n, S)
 @inline Base.:*(n::Number, S::AbstractTensor) = _map(x -> n*x, S)
 @inline Base.:/(S::AbstractTensor, n::Number) = _map(x -> x/n, S)
+
+# squaring method should be used for performance
 function Base.:^(S::SecondOrderTensor, p::Int)
     if p == 1
         return S
     elseif p == 0
         return one(S)
-    elseif p == -1
-        return inv(S)
     elseif p < 0
-        throw(DomainError())
+        return inv(S)^(-p)
     end
     t = S
     for _ in 2:p
--- a/test/test_misc.jl
+++ b/test/test_misc.jl
@@ -157,6 +157,15 @@ for T in (Float32, Float64), dim in (1,2,3), order in (1,2,4), TensorType in (Te
 
         @test (@inferred -t) == zero(t) - t
         @test isa(-t, $TensorType{$order, $dim})
+
+        if $order == 2
+            # Power by integer: ^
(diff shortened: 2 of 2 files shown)


KristofferC:
If we allowed for `*` we would get this atuomatically... The implementation here is a bit inefficient because they do not use the power by squaring method.

KeitaNakamura:
> If we allowed for * we would get this atuomatically...
Yes, I know but I thought you don't want to allow to use `*` operator. So, should I follow the original julia's code?

KristofferC:
I think this is fine. I don't expect large powers to be used much and we can fix it then if someone complains.

src/basic_operations.jl
```diff
@@ -18,6 +18,22 @@
...
 @inline Base.:*(n::Number, S::AbstractTensor) = _map(x -> n*x, S)
 @inline Base.:/(S::AbstractTensor, n::Number) = _map(x -> x/n, S)
+function Base.:^(S::SecondOrderTensor, p::Int)
+    if p == 1
+        return S
+    elseif p == 0
+        return one(S)
+    elseif p == -1
+        return inv(S)
+    elseif p < 0
```
KristofferC: Why throw domain error here? Can't you do `inv(S)^(-p)`?

KeitaNakamura: Oh, I would fix it. But might be `inv(S)^p`?

KeitaNakamura: No, sorry.

KeitaNakamura: BTW, how could I avoid errors in julia v0.6?

KristofferC: What's the problem on 0.6?

test/test_misc.jl
```diff
@@ -157,6 +157,15 @@ for T in (Float32, Float64), dim in (1,2,3), order in (1,2,4), TensorType in (Te
...
         @test (@inferred -t) == zero(t) - t
         @test isa(-t, $TensorType{$order, $dim})
+
+        if $order == 2
+            # Power by integer: ^
+            @test t^0 ≈ one(t)
+            @test t^1 ≈ t
+            @test t^2 ≈ t ⋅ t
+            @test t^3 ≈ t ⋅ t ⋅ t
+            @test t^-1 ≈ inv(t)
```
KristofferC: Could you add a test for `t^-2` as well

src/basic_operations.jl
KeitaNakamura: Oh, actually Travis would fail in "constructors" section, not in "simple math".

KeitaNakamura: The errors are following:

constructors: Test Failed
  Expression: eltype(tens_arr1) == (Tensors.Tensor){1, 1, Float64, N}
   Evaluated: Tensors.Tensor{1,1,T,M} where M where T<:Real == Tensors.Tensor{1,1,Float64,1}

KeitaNakamura: I would add more error message just in case.

Stacktrace:
 [1] macro expansion at /home/travis/.julia/v0.6/Tensors/test/test_misc.jl:42 [inlined]
 [2] macro expansion at ./test.jl:853 [inlined]
 [3] macro expansion at /home/travis/.julia/v0.6/Tensors/test/runtests.jl:9 [inlined]
 [4] macro expansion at /home/travis/.julia/v0.6/TimerOutputs/src/TimerOutput.jl:128 [inlined]
 [5] macro expansion at /home/travis/.julia/v0.6/Tensors/test/runtests.jl:8 [inlined]
 [6] anonymous at ./<missing>:?

fredrikekre:
I think we should do this with `Base.literal_pow` on julia06

KristofferC:
Some change in Base made tests fail..

fredrikekre:
Its this:

```julia
julia> zeros(Tensor{2, 3}, 3)
3-element Array{Tensors.Tensor{2,3,T,M} where M where T<:Real,1}:
 [0.0 0.0 0.0; 0.0 0.0 0.0; 0.0 0.0 0.0]
 [0.0 0.0 0.0; 0.0 0.0 0.0; 0.0 0.0 0.0]
 [0.0 0.0 0.0; 0.0 0.0 0.0; 0.0 0.0 0.0]
```

before `T` and `M` where correct here

KristofferC:
```
julia> @code_warntype zeros(Tensor{2, 3}, 3)

julia> 
```

wut?

KeitaNakamura:
Also

```julia
julia> methods(zeros, (Type{Tensor{2,3}}, Int))
# 0 methods for generic function "zeros":
```

KristofferC:
Bisected to https://github.com/JuliaLang/julia/pull/21183

PR #54, opened 2017-04-10
Fix zeros and ones

fredrikekre:
fix the test failure from #53

### DIFF
--- a/src/constructors.jl
+++ b/src/constructors.jl
@@ -103,8 +103,8 @@ Base.fill{T <: AbstractTensor}(el::Number, S::Type{T}) = apply_all(get_base(T), 
 Base.fill{T <: AbstractTensor}(f::Function, S::Type{T}) = apply_all(get_base(T), i -> f())
 
 # Array with zero/ones
-@inline Base.zeros{T <: AbstractTensor}(::Type{T}, dims...) = fill(zero(T), dims...)
-@inline Base.ones{T <: AbstractTensor}(::Type{T}, dims...) = fill(one(T), dims...)
+@inline Base.zeros{T <: AbstractTensor}(::Type{T}, dims::Int...) = fill!(Array{typeof(zero(T))}(dims), zero(T))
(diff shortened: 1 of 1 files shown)


fredrikekre:
Hmm actually it was enought to speify `Int`:

```diff
diff --git a/src/constructors.jl b/src/constructors.jl
index ab8ee10..ffed041 100644
--- a/src/constructors.jl
+++ b/src/constructors.jl
@@ -103,8 +103,8 @@ Base.fill{T <: AbstractTensor}(el::Number, S::Type{T}) = apply_all(get_base(T),
 Base.fill{T <: AbstractTensor}(f::Function, S::Type{T}) = apply_all(get_base(T), i -> f())
 
 # Array with zero/ones
-@inline Base.zeros{T <: AbstractTensor}(::Type{T}, dims...) = fill(zero(T), dims...)
-@inline Base.ones{T <: AbstractTensor}(::Type{T}, dims...) = fill(one(T), dims...)
+@inline Base.zeros{T <: AbstractTensor}(::Type{T}, dims::Int...) = fill(zero(T), dims...)
+@inline Base.ones{T <: AbstractTensor}(::Type{T}, dims::Int...) = fill(one(T), dims...)
 
 # diagm
 @generated function Base.diagm{T <: SecondOrderTensor}(S::Type{T}, v::Union{AbstractVector, Tuple})
```

KristofferC:
Thanks!

fredrikekre:
I think it must have taken another path, and made an `Array{T}` where `T = Tensor{2, 3, T, M}` instead of creating the array for the specific type, since `T` is not concrete

KristofferC:
Yeah, sounds reasonable.

fredrikekre:
I wonder if https://github.com/JuliaLang/julia/blob/master/base/array.jl#L250 should instead be 

```julia
$fname(T::Type, dims::Tuple) = fill!(Array{typeof($felt(T))}(Dims(dims)), $felt(T))
```
?

PR #55, opened 2017-04-10
use literal pow on 0.6 and make power typestable on 0.5

fredrikekre:
use the new `literal_pow` on 0.6 and restrict to powers of -1, 0 and 1.

Note that the implementation on master is type-unstable for other powers when used with a `SymmetricTensor`.

Did you have a real use case for other powers @KeitaNakamura ?

### DIFF
--- a/src/basic_operations.jl
+++ b/src/basic_operations.jl
@@ -19,22 +19,6 @@
 @inline Base.:*(n::Number, S::AbstractTensor) = _map(x -> n*x, S)
 @inline Base.:/(S::AbstractTensor, n::Number) = _map(x -> x/n, S)
 
-# squaring method should be used for performance
-function Base.:^(S::SecondOrderTensor, p::Int)
-    if p == 1
-        return S
-    elseif p == 0
-        return one(S)
-    elseif p < 0
-        return inv(S)^(-p)
-    end
-    t = S
-    for _ in 2:p
-        t = t ⋅ S
-    end
-    return t
-end
-
 Base.:+(S1::AbstractTensor, S2::AbstractTensor) = throw(DimensionMismatch("dimension and order must match"))
 Base.:-(S1::AbstractTensor, S2::AbstractTensor) = throw(DimensionMismatch("dimension and order must match"))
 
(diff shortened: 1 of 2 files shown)


KeitaNakamura:
I sometimes want to use it, for example, to compute invariants of tensors (see [https://en.wikipedia.org/wiki/Invariants_of_tensors](https://en.wikipedia.org/wiki/Invariants_of_tensors).)

KeitaNakamura:
Also for example [Saint Venant–Kirchhoff model](https://en.wikipedia.org/wiki/Hyperelastic_material). Of course I could write just `A ⋅ A`, but power is easier to read for me because the expression is the same as the equation.

fredrikekre:
Okay :) Will update the PR later, with a typestable version.

src/basic_operations.jl
```diff
@@ -48,3 +32,46 @@ end
...
+        S2 = S1
+        for i in 2:p
+            S2 = _powdot(S2, S1)
+        end
+        S2
+    end
+end
+
+@inline _powdot(S1::Tensor, S2::Tensor) = dot(S1, S2)
+@generated function _powdot{dim}(S1::SymmetricTensor{2, dim}, S2::SymmetricTensor{2, dim})
```
KristofferC: What does this do that is different from `dot`?

KristofferC: Aha, symmetrictensor.

KristofferC: So it is a type stable `dot(S, S)`?

fredrikekre: It will only be correct if `S1==S2` or if S1 is a power of `S2`

test/test_misc.jl
```diff
@@ -165,7 +165,7 @@ for T in (Float32, Float64), dim in (1,2,3), order in (1,2,4), TensorType in (Te
             @test t^2 ≈ t ⋅ t
             @test t^3 ≈ t ⋅ t ⋅ t
             @test t^-1 ≈ inv(t)
-            @test t^-2 ≈ inv(t)^2
+            @test t^-2 ≈ inv(t) ⋅ inv(t)
```
KristofferC: Perhaps sprinkle some `@inferred` since that is what the PR accomplish?

src/basic_operations.jl
fredrikekre: > So it is a type stable `dot(S, S)`?

Yes, but we can only use it with the above condition

test/test_misc.jl
fredrikekre: That won't parse correctly then for literal powers.

```julia
@inferred As^3
ERROR: use `⋅` (`\cdot`) for single contraction and `⊡` (`\boxdot`) for double contraction instead of `*`
Stacktrace:
 [1] power_by_squaring(::Tensors.SymmetricTensor{2,3,Float64,6}, ::Int64) at ./intfuncs.jl:182
 [2] ^(::Tensors.SymmetricTensor{2,3,Float64,6}, ::Int64) at ./linalg/dense.jl:326
```
but can add some `isa`

KeitaNakamura:
Thanks a lot!

PR #59, opened 2017-04-19
0.5 fix for AD

fredrikekre:
This fixes AD on 0.5, so lets merge this first and then rebase #58

### DIFF
--- a/src/automatic_differentiation.jl
+++ b/src/automatic_differentiation.jl
@@ -13,25 +13,25 @@ import ForwardDiff: Dual, partials, value
 # For dim = 1 the type is exactly the same.
 
 # Scalar
-@inline function _extract_value{N, T}(v::Dual{N, T}, ::Vec{N})
+@inline function _extract_value(v::Dual, ::Vec)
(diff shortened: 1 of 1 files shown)


KristofferC:
What is the problem with AD on 0.5 again?

fredrikekre:
Its only such that it will work with ForwardDiff master, where this happens now:
```julia
julia> v = rand(Vec{3}); gradient(norm, v)
3-element Tensors.Tensor{1,3,ForwardDiff.Dual{Void,Float64,3},3}:
 Dual{Void}(0.0,0.0,0.0,0.0)
 Dual{Void}(0.0,0.0,0.0,0.0)
 Dual{Void}(0.0,0.0,0.0,0.0)
```

fredrikekre:
Basically `Dual{N, T}` have changed to `Dual{Tag, T, N}`

KristofferC:
But 0.5 support was dropped from ForwardDiff for that change?

fredrikekre:
Right... Then #58 is enough

PR #68, opened 2017-06-01
fix a missed immutable -> struct

fredrikekre:


KristofferC:
:+1:

fredrikekre:
Travis timed out because of `using SIMD` on master.

KristofferC:
Perhaps should allow fail on nightly?

PR #76, opened 2017-09-21
Implement curl, div and laplace

fredrikekre:


fredrikekre:
```julia
julia> (∇ × f)(x)
3-element Tensors.Tensor{1,3,Float64,3}:
  0.835707 
 -0.789103 
 -0.0466032

julia> curl(f, x)
3-element Tensors.Tensor{1,3,Float64,3}:
  0.835707 
 -0.789103 
 -0.0466032
```

fredrikekre:
```julia
julia> (∇ ⋅ f)(x)
3.0

julia> div(f, x)
3.0
```

src/automatic_differentiation.jl
```diff
@@ -378,3 +379,73 @@ function hessian(f::F, v::Union{SecondOrderTensor, Vec}, ::Symbol) where {F}
...
+```jldoctest
+julia> f(x) = 2x;
+
+julia> x = rand(Vec{3});
+
+julia> div(f, x)
+6.0
+```
+"""
+Base.div(f::F, v::Vec) where {F} = trace(gradient(f, v))
```
fredrikekre: We should probably spell this out to `divergence` instead.

fredrikekre:
Failures are https://github.com/JuliaDiff/ForwardDiff.jl/pull/262#issuecomment-332171555

test/test_ad.jl
```diff
@@ -122,34 +122,73 @@ S(C) = S(C, μ, Kb)
...
+            # third derivatives
+            # @test laplace(x -> gradient(φ, x), x) ≈ gradient(x -> div(gradient(φ, x), x)) ≈ gradient(x -> laplace(φ, x), x)
+            @test laplace(x -> div(A, x), x) ≈ div(x -> gradient(x->div(A, x), x), x) # ≈ div(x -> laplace(A, x), x)
+            # @test laplace(x -> curl(A, x), x) ≈ -curl(x -> curl(curl(A, x), x), x) ≈ curl(x -> laplace(A, x))
+        end # loop T
     end # testsection
 end # testsection
+
+for i in 1:30
+    reset_timer!()
```
KristofferC: Just to check?

fredrikekre: yes....

fredrikekre:
Need to implement vector laplacian to enable the commented out tests some day too.

fredrikekre:
This is not gonna pass because of ForwardDiff. Passes locally, so ...

PR #77, opened 2017-09-26
don't import deprecated gradient

fredrikekre:


fredrikekre:
We can't do this until deprecations are removed actually.

src/automatic_differentiation.jl
```diff
@@ -1,5 +1,9 @@
 import ForwardDiff: Dual, partials, value
 
+if VERSION < v"1.0.0"
```
fredrikekre: The dream.

PR #78, opened 2017-10-03
vector laplacian as a broadcasting operation

fredrikekre:
Pretty cool, ref https://en.wikipedia.org/wiki/Vector_Laplacian

### DIFF
--- a/src/automatic_differentiation.jl
+++ b/src/automatic_differentiation.jl
@@ -398,7 +398,7 @@ julia> div(f, x)
 6.0
 ```
 """
-Base.div(f::F, v::Vec) where {F} = trace(gradient(f, v))
+Base.div(f::F, v::Vec) where {F<:Function} = trace(gradient(f, v))
 
 """
 ```julia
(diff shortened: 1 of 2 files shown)


src/automatic_differentiation.jl
```diff
@@ -433,19 +433,37 @@ curl(f::F, v::Vec{2, T}) where {F, T} = curl(f, Vec{3}((v[1], v[2], T(0))))
...
 ```
 """
 function laplace(f::F, v) where F
     return div(x -> gradient(f, x), v)
 end
 const Δ = laplace
+
+function Base.broadcast(::typeof(laplace), f::F, v::Vec{3}) where {F}
+    @inbounds begin
+        v1 = laplace(x->f(x)[1], v)
```
KristofferC: Isn't there a better way than calling the function 3 times?

fredrikekre: Probably.

fredrikekre:
```julia
function Base.broadcast(::typeof(laplace), f::F, v::Vec{3}) where {F}
    @inbounds begin
        v_dual = _load(_load(v))
        res = f(v_dual)
        v1 = res[1].partials[1].partials[1] + res[1].partials[2].partials[2] + res[1].partials[3].partials[3]
        v2 = res[2].partials[1].partials[1] + res[2].partials[2].partials[2] + res[2].partials[3].partials[3]
        v3 = res[3].partials[1].partials[1] + res[3].partials[2].partials[2] + res[3].partials[3].partials[3]
    end
    return Vec{3}((v1, v2, v3))
end
```

PR #82, opened 2017-11-19
Rotations

platawiec:
Added the following:

* Rotations for 2/4 order tensors in 3 dimensions (`rotate(x, u, angle)`)
* Generic rotations for 1/2/4 order tensors in n-dimensions (`rotate(x, u, v, angle)`). For this `u` and `v` are orthonormal vectors which define the n-2 dimensional space the rotation occurs in.
* Tests for 2-4 tensor dot product
* Tests for rotations in three dimensions

I tested out a generic fallback constructor which allowed n-dimensional tensors, but I may have been getting some test errors from that definition, or at least I wasn't sure where they were coming from. I'll look for a safer way to define such a constructor for a future PR.


KristofferC:
The test errors for the Voigt stuff look weird, perhaps they got broken with 0.6.1 or something... I'll try to look over this in the upcoming days. Many thanks for the PR.

PR #93, opened 2018-05-07
remove SIMD.jl

fredrikekre:
Probably not needed with llvm6 and SLP vectorization enabled by default.


KristofferC:
No explicit SIMD:

```
julia> a = rand(Tensor{4,3});

julia> @which a ⊡ a
dcontract(S1::FourthOrderTensor{dim}, S2::FourthOrderTensor{dim}) where dim
     @ Tensors ~/.julia/dev/Tensors/src/tensor_products.jl:139

julia> @btime $a ⊡ $a;
  80.764 ns (0 allocations: 0 bytes)
```

Explicit SIMD:
```
julia> @which a ⊡ a
dcontract(S1::Tensor{4, 3, T}, S2::Tensor{4, 3, T}) where T<:Union{Float16, Float32, Float64}
     @ Tensors ~/.julia/dev/Tensors/src/simd.jl:279

julia> @btime $a ⊡ $a;
  58.187 ns (0 allocations: 0 bytes)
```

PR #97, opened 2018-08-11
Set version to 1.0.0.

fredrikekre:


KristofferC:
What should we do with SIMD? I think it got tagged for 1.0. But now we have -O3 default. If we are not using SIMD should probably remove it as a dependency?

fredrikekre:
Well thats disappointing

PR #98, opened 2018-08-11
enable SIMD

KristofferC:
We should probably run a benchmark run vs 0.6 Tensors...


KristofferC:
Looks like this slightly changes the results in the doctests.

PR #99, opened 2018-09-14
define Vec on a tuple

KristofferC:
Its pretty common that you just want to convert a tuple to a Vec.


KristofferC:
Looks like `Vec(1)` infinite loops here.

PR #108, opened 2019-03-26
Fix return type of majorsymmetric(::SymmetricTensor).

fredrikekre:


fredrikekre:
Forgot to update the docstring

PR #109, opened 2019-03-28
Implement eigen decomposition for SymmetricTensor{4}.

fredrikekre:


test/test_misc.jl
```diff
@@ -325,6 +325,23 @@ for T in (Float32, Float64, F64), dim in (1,2,3)
 end
 end # of testset
 
+@testsection "eigen(::FourthOrderTensor)" begin
+for T in (Float32, Float64), dim in (1, 2, 3)
+    E = 200e9; ν = 0.3
+    λ = E*ν / ((1 + ν) * (1 - 2ν)); μ = E / (2(1 + ν))
+    δ(i,j) = i == j ? 1.0 : 0.0
+    f(i,j,k,l) = λ*δ(i,j)*δ(k,l) + μ*(δ(i,k)*δ(j,l) + δ(i,l)*δ(j,k))
```
KristofferC: Perhaps would be good with a matrix that has more non zeros?

PR #119, opened 2019-08-19
Upgrade Documenter to 0.23 and build docs on Julia 1.1.

fredrikekre:


KristofferC:
Or go for 1.2 straight away?

PR #124, opened 2019-10-26
Enable order kwarg for Voigt ordering.

fredrikekre:
You can now do something like
```julia
const abaqus = [1 4 6; 0 2 5; 0 0 3]
tovoigt(A; order=abaqus, offdiagscale=2)
```

cc @terofrondelius


TeroFrondelius:
looks really good. Thanks a lot.

PR #129, opened 2019-12-10
re-export `mean` from Statistics

KeitaNakamura:


KristofferC:
Seems reasonable. Thanks.

PR #136, opened 2020-06-22
Use Tuple instead of get_data

KeitaNakamura:
This PR is for easily getting tuple data from tensors.

Sometimes I need to use both of `Tensors` and `StaticArrays` packages because `Tensors` doesn't support non-square matrix. In this case, I need to write `Tensors.get_data` to get the tuple. But, I thought that it might be better to define `Tuple` instead of `get_data` as used in `StaticArrays`.

I think there is no confusion because we already have documentation to construct tensors from tuples. This is just reverse method for it. What do you think?

I know I can just define it by myself after `using Tensors`, but it is useful if we have it :)

### DIFF
--- a/src/Tensors.jl
+++ b/src/Tensors.jl
@@ -88,7 +88,7 @@ const NonSymmetricTensors{dim, T} = Union{Tensor{2, dim, T}, Tensor{4, dim, T}, 
 ##############################
 # Utility/Accessor Functions #
 ##############################
-get_data(t::AbstractTensor) = t.data
+Base.Tuple(t::AbstractTensor) = t.data
 
 @pure n_components(::Type{SymmetricTensor{2, dim}}) where {dim} = dim*dim - div((dim-1)*dim, 2)
 @pure function n_components(::Type{SymmetricTensor{4, dim}}) where {dim}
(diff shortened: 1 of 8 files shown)


KristofferC:
It isn't really clear to me what `Tuple(::SymmetricTensor)` should return.

KeitaNakamura:
For fourth-order symmetric tensors, it isn't clear much as you said. But for second-order tensor, it is said in the documentation as

> For symmetric tensors, the order of the numbers in the input tuple is column by column, starting at the diagonal.

```Julia
julia> SymmetricTensor{2,2}((1.0,2.0,3.0))
2×2 SymmetricTensor{2,2,Float64,3}:
 1.0  2.0
 2.0  3.0
```

Thus, we can get tuple with this PR:

```Julia
julia> A = SymmetricTensor{2,2}((1.0,2.0,3.0));

julia> Tuple(A)
(1.0, 2.0, 3.0)
```

Actually, before this PR:

```Julia
julia> A = SymmetricTensor{2,2}((1.0,2.0,3.0));

julia> Tuple(A)
(1.0, 2.0, 2.0, 3.0)
```

Isn't it reasonable to use `Tuple`?

PR #145, opened 2021-02-19
Update some links from org move, update doc deps.

fredrikekre:


KristofferC:
Need to update Julia version the docs run at?

PR #154, opened 2021-07-01
Allows use of symbolic tensors

jfbarthelemy:
These changes should allow to use symbolic numbers such as `Sym` from `SymPy.jl` within `Tensors.jl` .


KristofferC (approved):
Great. At some point, we should add a few tests with symbolic numbers to make sure this doesn't regress.

fredrikekre:
I added a simple test, thanks for the PR!

fredrikekre (approved):

PR #155, opened 2021-09-07
Construct from fully specified

KnutAM:
``` julia
a = rand(SymmetricTensor{2,3})
b = typeof(a)((i,j)-> i==j ? 1.0 : 0.0)
```
did not work before this change was added. 
However, for some unknown reason, 
``` julia
a = rand(Tensor{2,3})
b = typeof(a)((i,j)-> i==j ? 1.0 : 0.0)
```
still fails (also causing the new tests to fail)

### DIFF
--- a/src/Tensors.jl
+++ b/src/Tensors.jl
@@ -63,10 +63,7 @@ julia> Tensor{1,3,Float64}((1.0, 2.0, 3.0))
 """
 struct Tensor{order, dim, T, M} <: AbstractTensor{order, dim, T}
     data::NTuple{M, T}
-
-    # this is needed to make Vec{3, Float64}(f::Function) work properly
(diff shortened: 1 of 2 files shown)


KnutAM:
After modifying the constructor for Tensor, even 
``` julia
a = rand(Tensor{2,3})
b = typeof(a)((i,j)-> i==j ? 1.0 : 0.0)
```
works. 

Removed comment on line [67](https://github.com/Ferrite-FEM/Tensors.jl/blob/380ca525d792c12c0b51c192f0a8f94be0243fb7/src/Tensors.jl#L67) in Tensors.jl, as this should be sorted by the uncommented line [154](https://github.com/Ferrite-FEM/Tensors.jl/blob/master/src/Tensors.jl#L154)

PR #163, opened 2021-11-01
Voigt symmetric bool

KnutAM:
When converting a boolean vector to a symmetric tensor, using the Voigt format, the result is a Float64 tensor:
```julia
typeof(fromvoigt(SymmetricTensor{2,2}, rand(Float32,3).>0))
SymmetricTensor{2, 2, Float64, 3}

typeof(fromvoigt(SymmetricTensor{2,2}, rand(Float32,3)))
SymmetricTensor{2, 2, Float32, 3}

# No problem with full tensor:
typeof(fromvoigt(Tensor{2,2}, rand(Float32,4).>0))
Tensor{2, 2, Bool, 4}

```

The reason is division by `true`, which gives a Float64, see [voigt.jl#L179](https://github.com/Ferrite-FEM/Tensors.jl/blob/3ff7c906048fcd0ddcd7d3530dc94b08925407bc/src/voigt.jl#L179)

My suggestion here is to use `T(1/offdiagonalscale)` on L179 (and similar for 4th order). 
I could not see any performance change for regular Float64 input.

### DIFF
--- a/src/voigt.jl
+++ b/src/voigt.jl
@@ -176,7 +176,7 @@ Base.@propagate_inbounds function fromvoigt(TT::Type{<: SymmetricTensor{2, dim}}
     return TT(function (i, j)
             i > j && ((i, j) = (j, i))
             i == j ? (return v[offset + order[i, j]]) :
-                     (return v[offset + order[i, j]] / offdiagscale)
+                     (return v[offset + order[i, j]] * T(1 / offdiagscale) )
         end)
 end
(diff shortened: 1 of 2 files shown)


fredrikekre:
Seems fine. In `fromvoigt` if you use scaling you will usually input `1/2` (i.e. a float) and in that case `T` must be float, and thus the data too, so `T(1 / offdiagscale)` should always work I think.

Issue #164, opened 2021-11-01
Slow otimesu/otimesl for symmetric tensors


KnutAM:
The otimesu and otimesl are more than 10 times slower for symmetric tensors compared to full tensors.

`x, xs = (rand(T{2,3}) for T in (Tensor, SymmetricTensor));`
 command | time 
 --- | --- 
 `@btime otimes($xs, $xs);`   |  2.400 ns (0 allocations: 0 bytes)
 `@btime otimesu($xs, $xs);` |  249.734 ns (0 allocations: 0 bytes) 
`@btime otimesl($xs, $xs);`  |  250.269 ns (0 allocations: 0 bytes)
`@btime otimesl($x, $x);`|  11.512 ns (0 allocations: 0 bytes)
`@btime otimesu($x, $x);` | 9.810 ns (0 allocations: 0 bytes)

KnutAM:
Not sure why this happens, perhaps memory alignment. 
A potential solution could be to convert to full tensors first. Also here some strange timings though
```julia
julia> @btime Tensor{2,3}($xs);
  6.000 ns (0 allocations: 0 bytes)

julia> @btime +($xs, zero($x));
  1.800 ns (0 allocations: 0 bytes)
```

PR #165, opened 2021-11-03
Make otimesu and otimesl faster for symmetric tensors

KnutAM:
See #164 

## Timings 
Too large variance to determine if this has any negative effect when giving full tensors), but doesn't seem to have any significant effect... 

```julia
a, b = (rand(Tensor{2,3}) for _ in 1:2);
as, bs = (rand(SymmetricTensor{2,3}) for _ in 1:2);`
```

Command | Master | PR
--- | --- | ---
`@btime otimesu($a, $b);` | 13.627 ns | 10.611 ns | 
`@btime otimesu($as, $bs);` | 334.421 ns |  20.641 ns | 
`@btime otimesl($a, $b);` | 18.437 ns | 11.700 ns | 
`@btime otimesl($as, $bs);` | 343.071 ns | 25.451 ns |

### DIFF
--- a/src/tensor_products.jl
+++ b/src/tensor_products.jl
@@ -184,8 +184,8 @@ julia> otimesu(A, B)
 ```
 """
 @inline function otimesu(S1::SecondOrderTensor{dim}, S2::SecondOrderTensor{dim}) where {dim}
-    S1_ = Tensor{2,dim}(S1) # Convert to full tensor if symmetric to make 10x faster...
-    S2_ = Tensor{2,dim}(S2)
+    S1_ = convert(Tensor{2,dim}, S1) # Convert to full tensor if symmetric to make 10x faster... (see Tensors.jl#164)
+    S2_ = convert(Tensor{2,dim}, S2)
     return Tensor{4, dim}((i,j,k,l) -> S1_[i,k] * S2_[j,l])
 end
 
--- a/src/tensor_products.jl
+++ b/src/tensor_products.jl
@@ -184,8 +184,8 @@ julia> otimesu(A, B)
 ```
 """
 @inline function otimesu(S1::SecondOrderTensor{dim}, S2::SecondOrderTensor{dim}) where {dim}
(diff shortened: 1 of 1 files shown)


src/tensor_products.jl
```diff
@@ -184,7 +184,9 @@ julia> otimesu(A, B)
 ```
 """
 @inline function otimesu(S1::SecondOrderTensor{dim}, S2::SecondOrderTensor{dim}) where {dim}
-    return Tensor{4, dim}((i,j,k,l) -> S1[i,k] * S2[j,l])
+    S1_ = Tensor{2,dim}(S1) # Convert to full tensor if symmetric to make 10x faster...
+    S2_ = Tensor{2,dim}(S2)
```
fredrikekre, line 188:
```suggestion
    S1_ = convert(Tensor{2,dim}, S1) # Convert to full tensor if symmetric to make 10x faster... (see Tensors.jl#164)
    S2_ = convert(Tensor{2,dim}, S2)
```

src/tensor_products.jl
```diff
@@ -218,7 +220,9 @@ julia> otimesl(A, B)
 ```
 """
 @inline function otimesl(S1::SecondOrderTensor{dim}, S2::SecondOrderTensor{dim}) where {dim}
-    return Tensor{4, dim}((i,j,k,l) -> S1[i,l] * S2[j,k])
+    S1_ = Tensor{2,dim}(S1) # Convert to full tensor if symmetric to make 10x faster...
+    S2_ = Tensor{2,dim}(S2)
```
fredrikekre, line 224:
```suggestion
    S1_ = convert(Tensor{2,dim}, S1) # Convert to full tensor if symmetric to make 10x faster... (see Tensors.jl#164)
    S2_ = convert(Tensor{2,dim}, S2)
```

KnutAM:
Nice, that should be guaranteed to remove any overhead!

fredrikekre:
No SIMD and lots of data shuffling in the slow case.

PR #172, opened 2021-12-21
Correction of bug in minorsymmetric

jfbarthelemy:
The right formula for "minorsymmetrization" is `(Sᵢⱼₖₗ+Sⱼᵢₖₗ+Sᵢⱼₗₖ+Sⱼᵢₗₖ)/4` and not `(Sᵢⱼₖₗ+Sⱼᵢₖₗ+Sᵢⱼₖₗ+Sᵢⱼₗₖ)/4`

### DIFF
--- a/src/symmetric.jl
+++ b/src/symmetric.jl
@@ -38,7 +38,7 @@ Compute the minor symmetric part of a fourth order tensor, return a `SymmetricTe
             @inbounds if i == j && k == l
                 return S[i,j,k,l]
             else
-                return (S[i,j,k,l] + S[j,i,k,l] + S[i,j,k,l] + S[i,j,l,k]) / 4
+                return (S[i,j,k,l] + S[j,i,k,l] + S[i,j,l,k] + S[j,i,l,k]) / 4
             end
         end
     )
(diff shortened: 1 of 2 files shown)


KnutAM:
Nice, this would introduce hard to find bugs! I would think a test to catch this would be nice to add, what about
```julia
for dim=1:3
a,b=[rand(Tensor{2,dim}) for _ in 1:2];
as,bs = symmetric.([a,b]);
@test minorsymmetric(a⊗b)≈as⊗bs
end
```
?

jfbarthelemy:
> Nice, this would introduce hard to find bugs! I would think a test to catch this would be nice to add, what about
> 
> ```julia
> for dim=1:3
> a,b=[rand(Tensor{2,dim}) for _ in 1:2];
> as,bs = symmetric.([a,b]);
> @test minorsymmetric(a⊗b)≈as⊗bs
> end
> ```
> 
> ?

jfbarthelemy:
> Nice, this would introduce hard to find bugs! I would think a test to catch this would be nice to add, what about
> 
> ```julia
> for dim=1:3
> a,b=[rand(Tensor{2,dim}) for _ in 1:2];
> as,bs = symmetric.([a,b]);
> @test minorsymmetric(a⊗b)≈as⊗bs
> end
> ```
> 
> ?

jfbarthelemy:
> Nice, this would introduce hard to find bugs! I would think a test to catch this would be nice to add, what about
> 
> ```julia
> for dim=1:3
> a,b=[rand(Tensor{2,dim}) for _ in 1:2];
> as,bs = symmetric.([a,b]);
> @test minorsymmetric(a⊗b)≈as⊗bs
> end
> ```
> 
> ?

Sure. By the way the test you propose doesn't pass with the old version and passes with the new one.

fredrikekre:
Thank you!

Issue #174, opened 2022-01-07
Possible bug when computing derivatives of principal directions with AD


victorjesusamoresmedianero:
We found that there is some inconsistency when trying to compute the derivative of the eigvecs dyadic product wrt. a symmetric tensor. To support this claim a comparison with an analytical expression for the derivative is used: 

if we have a second order symmetric tensor $\bm{A}$ and its principal directions (eigenvector) are $\bm{l}_1$, $\bm{l}_2$, $\bm{l}_3$ we would like to compute the derivative \dfrac{\partial \bm{l}_i \otimes \bm{l}_i}{\partial \bm{A}}

Code to compare:
```julia
using ForwardDiff
Base.exponent(VD::ForwardDiff.Dual) = exponent(VD.value)
Base.precision(::Type{TT}) where{TT<:ForwardDiff.Dual{T,V}} where{T,V} = precision(V)

function M11f(A)
    l = eigvecs(A)
    return symmetric(l[:,1] ⊗ l[:,1])
end

function dM11fAnalytic(A)
    λ = eigvals(A)
    l = eigvecs(A)
    𝕀s = one(SymmetricTensor{4,3})
    M11 = symmetric(l[:,1] ⊗ l[:,1])
    M22 = symmetric(l[:,2] ⊗ l[:,2])
    M33 = symmetric(l[:,3] ⊗ l[:,3])
    
    ∂M11∂A = (otimesu(M11,M22) + otimesu(M22,M11))/(λ[1]-λ[2]) + (otimesu(M11,M33) + otimesu(M33,M11))/(λ[1]-λ[3])

   return 𝕀s⊡(∂M11∂A⊡𝕀s)
end

A0 = SymmetricTensor{2,3}((1.2, 0., 0., 0.9, 0., 0.75))

gradient(M11f,A0)
dM11fAnalytic(A0)
```

Results:

```julia
julia> dM11fAnalytic(A0)[:,:,1,3]
3×3 Matrix{Float64}:
 -0.0      -0.0  -1.11111
 -0.0      -0.0  -0.0
 -1.11111  -0.0  -0.0

julia> gradient(M11f,A0)[:,:,1,3]
3×3 Matrix{Float64}:
 0.0  0.0  0.0
 0.0  0.0  0.0
 0.0  0.0  0.0

julia> dM11fAnalytic(A0)[:,:,2,3]
3×3 Matrix{Float64}:
 -0.0  -0.0      -0.0
[...]

PR #175, opened 2022-01-09
WIP solving #174

KnutAM:
Initial work on solving problem with ad for eigenvalues #174

### DIFF
--- a/src/Tensors.jl
+++ b/src/Tensors.jl
@@ -4,6 +4,8 @@ module Tensors
 
 import Base.@pure
 
+import ForwardDiff: Dual, partials, value, Tag
+
 import Statistics
 using Statistics: mean
 using LinearAlgebra
--- a/src/automatic_differentiation.jl
+++ b/src/automatic_differentiation.jl
@@ -1,5 +1,3 @@
-import ForwardDiff: Dual, partials, value, Tag
-
 @static if isdefined(LinearAlgebra, :gradient)
     import LinearAlgebra.gradient
 end
--- a/src/eigen.jl
+++ b/src/eigen.jl
@@ -1,3 +1,7 @@
+# Support functions
+get_max_iterations(::Type{T}) where{T} = 2 * (1 + precision(T) - exponent(floatmin(T)))
+get_max_iterations(::Type{T}) where{T<:Dual{TT,VT}} where{TT,VT} = get_max_iterations(VT)
+
 # MIT License: Copyright (c) 2016: Andy Ferris.
 # See LICENSE.md for further licensing test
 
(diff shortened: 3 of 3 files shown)


KnutAM:
To start added internal `get_max_iterations` to make code run when called with dual number.
Does not resolve the correctness of the solution though. But 
```julia
using Tensors

function M11f(A)
    l = eigvecs(A)
    return symmetric(l[:,1] ⊗ l[:,1])
end

A0 = SymmetricTensor{2,3}((1.2, 0., 0., 0.9, 0., 0.75))
B0 = rand(SymmetricTensor{2,3})

Δ = rand(SymmetricTensor{2,3})*1.e-6

println("A0: ", M11f(A0+Δ) ≈ M11f(A0) + gradient(M11f,A0)⊡Δ)
println("B0: ", M11f(B0+Δ) ≈ M11f(B0) + gradient(M11f,B0)⊡Δ)
```
produces 
```julia
A0: false
B0: true
```
Seems like the wrong result is caused by a special case given by the A0 tensor in #174. 
When debugging, no iterations are necessary for A0. My guess here is that we cannot expect AD to give the correct result for an iterative procedure if the input is such that no iterations occur? Perhaps derivatives must be specifically implemented?

KnutAM:
Based on Kristoffer's AD talk today, I think [`get_cos_sin()` in eigen.jl](https://github.com/Ferrite-FEM/Tensors.jl/blob/c3b72b3101fcf1389e6eab77e7dec3370133a822/src/eigen.jl#L74-L88) could be to blame:
```julia
    function get_cos_sin(u::T,v::T) where {T}
        max_abs = max(abs(u), abs(v))
        if max_abs > 0
            u,v = (u,v) ./ max_abs
            len = sqrt(u^2 + v^2)
            cs, sn = (u,v) ./ len
            if cs > 0
                cs = -cs
                sn = -sn
            end
            T(cs), T(sn)
        else
            T(-1), T(0)
        end
    end
```
The problem is that `len=0` (or at least `real(len))` when we need to force it to go via the calculations with a Dual number, so we get division by zero. I don't have a solution on top of my head right now for this...

KnutAM:
> ```julia 
> function get_cos_sin(u, v)
>     α = atan(v, u) + π
>     return cos(α), sin(α)
> end
> ```

Tested this with [ForwardDiff's NaN-safe mode](https://juliadiff.org/ForwardDiff.jl/latest/user/advanced.html#Fixing-NaN/Inf-Issues-1), but that did not work.

Note to possible future implementation: https://github.com/JuliaDiff/ForwardDiff.jl/issues/111#issuecomment-636928169

KnutAM:
Solved by #182 and [ForwardDiff.jl-PR575](https://github.com/JuliaDiff/ForwardDiff.jl/pull/575)

PR #176, opened 2022-01-14
Enable insertion of known derivative in autodiff chain

KnutAM:
This PR considers creating a mechanism for inserting a known derivative into a function that is part of a chain of functions differentiated using the `Tensors.jl`'s `gradient` method. ~~The new method is `insert_gradient`, and `extract_value` has also been exported.~~ Using the suggestion from @KristofferC below, it introduces the macro `@implement_gradient` 

The main goal is to be able to provide analytical derivatives when automatic differentiation fails, see e.g. #174 .
As a side benefit, if a known analytical derivative of part of a function that is slow to call using dual numbers, this method can be used to provide a more efficient implementation (not the case for most functions in `Tensors.jl` though)

Updated example (cases in a comment below are outdated, see also docs):
```julia
using Tensors
# The function to be given a known gradient
f(x) = x⋅x
# Define function to calculate both value and analytical derivative
function f_dfdx(x::Tensor{2,dim}) where{dim}
    fval = f(x)
    I2 = one(Tensor{2,dim})
    df = otimesu(I2, transpose(x)) + otimesu(x, I2)
    return fval, df
end
# Connect the analytical derivative for `f` to `f_dfdx`
@implement_gradient f f_dfdx
```

Suggestions to improve the method are very welcome! 

~~In order to work for #174 / #175 specifically, it seems to require support for 3rd order tensors if we don't want an ad-hoc fix for the specific differentiation case (because we differentiate an (eigen)vector wrt. to a 2nd order at some point in the differentiation chain). I guess supporting 3rd order should be a separate PR though~~

#174 can now be solved with this PR if the user manually provides the derivative for their specific function. This PR might be used to solve AD of `Tensors.jl`'s eigenvector calculation, but requires dealing with 3rd order tensors and is not considered here. 

#### Help wanted
- [x] Decide if functionality (`_insert_gradient`) should be API or remain internal: @fredrikekre suggested to make public.
- ~~Potential efficiency gains? (in some cases much slower execution time, but also sometimes faster)~~: OK for now

#### Remaining tasks
- [x] Support symmetric tensors
- [x] Add tests
- [x] Add documentation

### DIFF
--- a/src/automatic_differentiation.jl
+++ b/src/automatic_differentiation.jl
@@ -251,7 +251,7 @@ function insert_gradient(f::Number, dfdg::SecondOrderTensor{dim}, g::SecondOrder
     return Dual{Tg}(f, get_data(dfdx))
 end
 
-function _insert_gradient(f::Vec{dim}, dfdg::Tensor{2,dim}, g::Vec{dim, <:Dual{Tg}}) where{dim,Tg}
+function insert_gradient(f::Vec{dim}, dfdg::Tensor{2,dim}, g::Vec{dim, <:Dual{Tg}}) where{dim,Tg}
     fdata = get_data(f)
     dgdx = _extract_gradient(g, _get_original_gradient_input(g))
     dfdx = dfdg ⋅ dgdx
@@ -261,7 +261,7 @@ function _insert_gradient(f::Vec{dim}, dfdg::Tensor{2,dim}, g::Vec{dim, <:Dual{T
     return y
 end
 
-function _insert_gradient(f::SecondOrderTensorN{dim, <:Any, N}, dfdg::FourthOrderTensor{dim}, g::SecondOrderTensorN{dim,<:Dual{Tg},N}) where{dim,Tg,N}
+function insert_gradient(f::SecondOrderTensorN{dim, <:Any, N}, dfdg::FourthOrderTensor{dim}, g::SecondOrderTensorN{dim,<:Dual{Tg},N}) where{dim,Tg,N}
     fdata = get_data(f)
     dgdx = _extract_gradient(g, _get_original_gradient_input(g))
     dfdx = dfdg ⊡ dgdx
--- a/src/automatic_differentiation.jl
+++ b/src/automatic_differentiation.jl
@@ -225,7 +225,7 @@ function _insert_gradient(f::Number, ∇f::Tensor{2,dim}, x::Tensor{2,dim,<:Dual
     diffdata = get_data(∇f ⊡ _extract_gradient(x, prev_input_for_type_info))
     dim2 = dim^2
     @inbounds begin
-    y = Dual{Tg}(f, ntuple(j->diffdata[j],dim2))
+    y = Dual{Tg}(f, diffdata)
     end
     return y
 end
--- a/src/automatic_differentiation.jl
+++ b/src/automatic_differentiation.jl
@@ -190,29 +190,19 @@ end
 # Gradient insertion #
 ######################
 
-# Inserters are supposed to take a tensor of real values and convert it
-# into a tensor of dual values where the seeds are correctly defined.
-
 # Insertions get the real value and derivative of a function, as well 
 # a tensor of dual values that was the initial input to that function. 
 # A new tensor of dual values are then created, to emulate the function
 # being run with dual numbers (i.e. inserting the analytical gradient)
-# As opposed to with gradient extraction, we cannot know the input given 
-# to the previous function, except by inspection of the input tensor with 
-# dual entries: The number of partials (of the entry) together with order 
-# and dimension of the input tensor allow us to deduce this for a regular 
-# tensor. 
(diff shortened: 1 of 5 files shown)


KnutAM:
#### Test cases
```julia
import Tensors: Dual, _insert_gradient, _extract_value
using Tensors, ForwardDiff
using BenchmarkTools
using Test

if !isdefined(Main, :DualTensor)
    const DualTensor=Dual{<:ForwardDiff.Tag{<:Any,<:AbstractTensor}}
end

g1(x)=x
f1(x) = g1(x)

g2(x) = x⋅x
f2(x) = g2(x)

g3(x) = x⊡x
f3(x) = g3(x)

function f1(x::Tensor{2,dim,T}) where{dim, T<:DualTensor}
    fval = f1(_extract_value(x))
    ∇f = one(Tensor{4,dim})
    return _insert_gradient(fval, ∇f, x)
end

function f2(x::Tensor{2,dim,T}) where{dim, T<:DualTensor}
    xval = _extract_value(x)
    fval = f2(xval)
    I2 = one(Tensor{2,dim})
    ∇f = otimesu(I2, transpose(xval)) + otimesu(xval, I2)
    return _insert_gradient(fval, ∇f, x)
end

function f3(x::Tensor{2,dim,T}) where{dim, T<:DualTensor}
    xval = _extract_value(x)
    fval = f3(xval)
    ∇f = 2*xval
    return _insert_gradient(fval, ∇f, x)
end

for dim in 1:3
    a = rand(Tensor{2,dim})
    for (f,g) in ((f1,g1),(f2,g2),(f3,g3))
        println("f = $f, dim = $dim")
        @test gradient(f,a)==gradient(g,a)
        print("analytical: "); @btime gradient($f, $a); 
        print("autodiff:   "); @btime gradient($g, $a);
        println("")
    end
end
```

#### Results
```julia
f = f1, dim = 1
analytical:   0.800 ns (0 allocations: 0 bytes)
autodiff:     0.800 ns (0 allocations: 0 bytes)

f = f2, dim = 1
analytical:   6.900 ns (0 allocations: 0 bytes)
autodiff:     0.800 ns (0 allocations: 0 bytes)

f = f3, dim = 1
analytical:   0.800 ns (0 allocations: 0 bytes)
autodiff:     0.800 ns (0 allocations: 0 bytes)

f = f1, dim = 2
analytical:   13.126 ns (0 allocations: 0 bytes)
autodiff:     1.000 ns (0 allocations: 0 bytes)

f = f2, dim = 2
analytical:   23.571 ns (0 allocations: 0 bytes)
autodiff:     17.435 ns (0 allocations: 0 bytes)

f = f3, dim = 2
analytical:   5.000 ns (0 allocations: 0 bytes)
autodiff:     11.400 ns (0 allocations: 0 bytes)

f = f1, dim = 3
analytical:   101.067 ns (0 allocations: 0 bytes)
autodiff:     5.600 ns (0 allocations: 0 bytes)

f = f2, dim = 3
analytical:   132.686 ns (0 allocations: 0 bytes)
autodiff:     81.186 ns (0 allocations: 0 bytes)

f = f3, dim = 3
analytical:   29.335 ns (0 allocations: 0 bytes)
autodiff:     34.211 ns (0 allocations: 0 bytes)
```

KristofferC:
Would be good with a few tests but otherwise this looks ok. It is a bit awkward to special case the function on `Dual`.  A possible API could be similar to what I did in https://github.com/JuliaDiff/ForwardDiff.jl/pull/165. But this is ok as well.

KnutAM:
Cool, I'll try to add a few tests tomorrow!
Do you mean to provide a macro like `@implement_gradient`? I think that looks nice! 
As a novice Julia programmer, however, I'm a bit concerned that users (like me) can find debugging harder when using macros, but perhaps that is unwarranted?

KnutAM:
I'm not sure if it is ok that I included allowing the additional operations for the [open product](https://github.com/Ferrite-FEM/Tensors.jl/pull/176/files#diff-6c59b8aa0954bb10a4d33071d00cbd60b94e53098bf39f922082d162537c4564R131-R133), but it allowed me to write fewer functions and I think it makes sense?

src/automatic_differentiation.jl
```diff
@@ -20,6 +20,14 @@ end
 ####################
 
 # Scalar output -> Scalar value
+"""
+    function _extract_value(v::ForwardDiff.Dual)
```
KristofferC, line 24: If this (and `_insert_gradient`) are documented, perhaps remove the underscore?

KristofferC (approved):
LGTM, nice job!

src/automatic_differentiation.jl
KnutAM: They are documented, but I didn't include them in the docs.
(As I documented them before using your @implement_gradient idea). 
Should I then remove underscore, include in docs, and export them, or just remove underscore?

KristofferC:
Let's go with this for now :)

PR #177, opened 2022-02-11
Fix outdated docs on symmetrization of 4th order tensors

KnutAM:


KnutAM:
Is this PR ok @fredrikekre?

PR #180, opened 2022-03-25
Eigen: Use Mandel representation for 4th order eigen

fredrikekre:
This lets us dispatch to LAPACKs symmetric eigenvalue solver, and also
makes sure that the returned eigentensors are normalized.


KnutAM:
What if `a=rand(SymmetricTensor{2,2}); b=rand(SymmetricTensor{2,2}); M=tomandel(a⊗b)`
Then `M != Hermitian(M)`?

fredrikekre:
But then it isn't diagonalizable, right? But should add the check instead of truncating it to be Hermitian.

KnutAM:
True, the eigenvalues might be complex but still defined. 
But a check would be nice and perhaps then document that only major symmetric fourth order tensors are supported?

PR #185, opened 2022-09-23
Export rotation_tensor and add implementation for 2D

fredrikekre:
This patch exports the existing rotation_matrix function, and adds a new method rotation_matrix(x::Number) which returns the two-dimensional rotation matrix for rotation x radians around the out-of-plane axis.


KnutAM:
Would it make sense to rename it to `rotation_tensor` instead before exporting and documenting?

fredrikekre:
Yea sounds good.

PR #195, opened 2023-03-07
Add orthogonal and isparallel function

lijas:
I have wanted this a couple of times :)

### DIFF
--- a/src/math_ops.jl
+++ b/src/math_ops.jl
@@ -378,3 +378,32 @@ function rotate(x::SymmetricTensor{4}, args...)
     R = rotation_tensor(args...)
     return unsafe_symmetric(otimesu(R, R) ⊡ x ⊡ otimesu(R', R'))
 end
+
+isparallel(v1::Vec{3}, v2::Vec{3}) = isapprox(Tensors.cross(v1,v2), zero(v1), atol = 1e-14)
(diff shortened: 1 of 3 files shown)


src/math_ops.jl
```diff
@@ -378,3 +378,32 @@ function rotate(x::SymmetricTensor{4}, args...)
     R = rotation_tensor(args...)
     return unsafe_symmetric(otimesu(R, R) ⊡ x ⊡ otimesu(R', R'))
 end
+
+isparallel(v1::Vec{3}, v2::Vec{3}) = isapprox(Tensors.cross(v1,v2), zero(v1), atol = 1e-14)
```
KnutAM, line 382: Can we get rid of the absolute tolerance by doing this (which also generalizes to second order)
```suggestion
isparallel(a::Vec, b::Vec) = abs(cdot(a,b)) ≈ norm(a)*norm(b)
isparallel(a::SecondOrderTensor, b::SecondOrderTensor) = abs(dcontract(a,b)) ≈ norm(a)*norm(b)
```

KnutAM:
Cool!
Perhaps also the orthogonal can be generalized to 2nd order tensors?

src/math_ops.jl
fredrikekre: `≈` also uses a tolerance though, right?

dkarrasch: This cries for a user-defined tolerance with some default value. Think of it conversely: any tolerance describes a cone around one vector in which all vectors are considered as parallel to the reference vector. Same with orthogonality.

PR #198, opened 2023-03-31
3D rotation by Euler angles

koehlerson:


src/math_ops.jl
```diff
@@ -346,6 +346,26 @@ function rotation_tensor(θ::Number)
     return Tensor{2, 2}((c, s, -s, c))
 end
 
+@doc raw"""
+    rotation_tensor(ψ::Number,θ::Number,ϕ::Number)
+
```
termi-official, line 351: While I appreciate the feature, can we leave a note for the user that this might be hamful to use in specific scenarios (due to the possibility to run into gimbal locking)?

termi-official: I think this should be here because from my experience many users are not aware of this.

fredrikekre:
@KnutAM wants a reference in the docstring since there are apparently multiple definitions of Euler angles :)

koehlerson:
http://eecs.qmul.ac.uk/~gslabaugh/publications/euler.pdf I got it from here, is it formal enough or do you want a textbook ref @KnutAM ?

koehlerson:
Its pretty much the same notation as the german wiki site https://de.wikipedia.org/wiki/Eulersche_Winkel#Abbildungsmatrix_(aktive_Drehung)

KnutAM:
I was just reading on wikipedia: https://en.wikipedia.org/wiki/Euler_angles
Seems to be six possibilities: https://en.wikipedia.org/wiki/Euler_angles#Signs,_ranges_and_conventions for proper Euler angles?

KnutAM:
I just think it would be nice to have a reference (or even preferably describe directly the definitions of the input angles) in the docstring to the definition used. The sequential rotations on the German wiki would be a great note IMO.

src/math_ops.jl
```diff
@@ -346,6 +346,29 @@ function rotation_tensor(θ::Number)
...
+Return the three-dimensional rotation matrix corresponding to the rotation described
+by the three Euler angles $ψ, θ, ϕ$.
+
+```math
+R(ψ,θ,ϕ) = R_x(ψ)R_y(θ)R_z(ϕ)
+```
+
```
KnutAM, line 358:
```suggestion
see e.g. [here](http://eecs.qmul.ac.uk/~gslabaugh/publications/euler.pdf) for a complete description.

```

src/math_ops.jl
```diff
@@ -346,6 +346,29 @@ function rotation_tensor(θ::Number)
...
+Note that the [gimbal lock phenomena](https://en.wikipedia.org/wiki/Gimbal_lock) can occur when using
+this rotation tensor parametrization.
+"""
+function rotation_tensor(ψ::Number,θ::Number,ϕ::Number)
+    # See http://eecs.qmul.ac.uk/~gslabaugh/publications/euler.pdf
```
KnutAM, line 363:
```suggestion
```

KnutAM (approved):
Sorry, I must have been pretty tired yesterday, it was already pretty clear @koehlerson !
Could potentially move the link to the docstring.

src/math_ops.jl
```diff
@@ -346,6 +346,29 @@ function rotation_tensor(θ::Number)
     return Tensor{2, 2}((c, s, -s, c))
 end
 
+@doc raw"""
+    rotation_tensor(ψ::Number,θ::Number,ϕ::Number)
+
+Return the three-dimensional rotation matrix corresponding to the rotation described
+by the three Euler angles $ψ, θ, ϕ$.
```
fredrikekre, line 353:
```suggestion
by the three Euler angles ``ψ``, ``θ``, ``ϕ``.
```

src/math_ops.jl
```diff
@@ -346,6 +346,29 @@ function rotation_tensor(θ::Number)
     return Tensor{2, 2}((c, s, -s, c))
 end
 
+@doc raw"""
```
fredrikekre, line 349:
```suggestion
"""
```

src/math_ops.jl
```diff
@@ -346,6 +346,29 @@ function rotation_tensor(θ::Number)
...
+    rotation_tensor(ψ::Number,θ::Number,ϕ::Number)
```
fredrikekre, line 350:
```suggestion
    rotation_tensor(ψ::Number, θ::Number, ϕ::Number)
```

PR #205, opened 2023-08-11
3rd order tensors

KnutAM:
Start of limited support for 3rd order tensors.
Currently, the aim is just to implement what is required for mapping of Hcurl and Hdiv interpolations in Ferrite (Co- and Contra-variant Piola mapping)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
-## [1.15.1]
+## [1.16.0]
+
 ### Added
  - Partial support for 3rd order Tensors [#205][github-205]
     * All construction methods, e.g. `zero(Tensor{3})`, `rand(Tensor{3})`, `Tensor{3}((i,j,k)->f(i,j,k))`
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,0 +1,25 @@
+# Tensors changelog
+
+All notable changes to this project will be documented in this file.
+
+The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
+and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
+
+## Unreleased
+
+## [1.15.1]
+### Added
+ - Partial support for 3rd order Tensors [#205][github-205]
+    * All construction methods, e.g. `zero(Tensor{3})`, `rand(Tensor{3})`, `Tensor{3}((i,j,k)->f(i,j,k))`
+    * Gradient of 2nd order tensor wrt. vector
+    * `rotate(::Tensor{3})`
+    * `dcontract(::Tensor{D1}, ::Tensor{D2})` for (D1,D2) in ((2,3), (3,2), (3,4), (4,3))
+    * `otimes(::Vec, ::SecondOrderTensor)` and `otimes(::SecondOrderTensor, ::Vec)`
+    * `dot(::Tensor{D1}, ::Tensor{D2})` for (D1,D2) in ((3,1), (1,3), (2,3), (3,2))
+
+<!-- Release links -->
+[Unreleased]: https://github.com/Ferrite-FEM/Ferrite.jl/compare/v1.15.1...HEAD
+[1.15.1]: https://github.com/Ferrite-FEM/Ferrite.jl/compare/v1.15.0...v1.15.1
+
+<!-- GitHub pull request/issue links -->
+[github-205]: https://github.com/Ferrite-FEM/Tensors.jl/pull/205
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,8 +18,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
     * `dot(::Tensor{D1}, ::Tensor{D2})` for (D1,D2) in ((3,1), (1,3), (2,3), (3,2))
 
 <!-- Release links -->
-[Unreleased]: https://github.com/Ferrite-FEM/Ferrite.jl/compare/v1.15.1...HEAD
-[1.15.1]: https://github.com/Ferrite-FEM/Ferrite.jl/compare/v1.15.0...v1.15.1
+[Unreleased]: https://github.com/Ferrite-FEM/Tensors.jl/compare/v1.15.1...HEAD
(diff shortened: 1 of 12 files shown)


src/automatic_differentiation.jl
termi-official, line 1: Vec input, Tensor{2} output?

KnutAM: Sorry, what do you mean? (I also implemented this by mistake, didn't actually need it😅)

KnutAM:
I think this is ready for review, https://github.com/Ferrite-FEM/Ferrite.jl/pull/798 relies on it. 

Would be good with some information in the docs about the limited/experimental support for 3rd order, since only a small subset of all methods are implemented. Any idea where is a good place @fredrikekre?

src/automatic_differentiation.jl
termi-official: `@inline function _extract_gradient(v::Vec, ::Tensor{2})`

KnutAM: I'm sorry if I misunderstood here, not sure if you want the reverse case implemented or think the notation is wrong.

The comment is because the input argument to the differentiated function was a `Vec` (signaled by the 2nd argument), just for dispatch, and the output of calling that function with a `Vec` is a `Tensor{2}` given as the first argument. This ordering and comments should follow the other places in the file i think.

termi-official: Ah sorry, I misread it here what input and output are. :D 

But yes, for completeness we can also think about implementing the other case, when we have Tensor as an input and Vec as the output.

KnutAM: Aha, ok. 

Yes, for sure that can be nice - but for now I think I want to leave this as partial support. 
If needed later, it is mostly mechanical work to implement, for me what takes most time is to develop the test cases that ensure that the implementation is correct. 

After this PR, we could open a tracking issue for 3rd-order tensor support.

fredrikekre:
The Changelog would be a start, but I see we don't have one here 🙂

src/math_ops.jl
```diff
@@ -21,6 +21,19 @@ julia> norm(A)
 @inline LinearAlgebra.norm(v::Vec) = sqrt(dot(v, v))
 @inline LinearAlgebra.norm(S::SecondOrderTensor) = sqrt(dcontract(S, S))
 
+@generated function LinearAlgebra.norm(S::Tensor{3,dim}) where {dim}
```
fredrikekre, line 24: I know this is copied from below, but I wonder how it compares to just delegating to StaticArrays.

KnutAM: No major difference, but 5-10 % faster with Tensors' implementation on my system...

```julia
t = rand(Tensor{3,3});
snorm(t::Tensor) = norm(SVector(Tensors.get_data(t)))
tnorm(t::Tensor) = norm(Tensors.get_data(t))

julia> @btime norm($t);
  8.800 ns (0 allocations: 0 bytes) # Run 1
  9.500 ns (0 allocations: 0 bytes) # Run 2

julia> @btime snorm($t);
  10.000 ns (0 allocations: 0 bytes) # Run 1
  10.000 ns (0 allocations: 0 bytes) # Run 2

julia> @btime tnorm($t);
  87.137 ns (0 allocations: 0 bytes) # 🥱
```

KnutAM: But if we really care about performance, and that would also reduce the amount of generated code...
```julia
julia> function _norm(t::NTuple{N,T}) where {N,T<:AbstractFloat}
           n = zero(T)
           @simd for v in t
               n += v^2
           end
           return sqrt(n)
       end
_norm (generic function with 1 method)

julia> _norm(t::Tensor) = _norm(Tensors.get_data(t))
_norm (generic function with 2 methods)

julia> @btime _norm($t3);
  4.000 ns (0 allocations: 0 bytes)
```

CHANGELOG.md
```diff
@@ -0,0 +1,25 @@
+# Tensors changelog
+
+All notable changes to this project will be documented in this file.
+
+The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
+and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
+
+## Unreleased
+
+## [1.15.1]
```
fredrikekre, line 10:
```suggestion
## [1.16.0]

```

fredrikekre (approved):

src/tensor_products.jl
```diff
@@ -68,6 +68,74 @@ end
...
+        ex1 = Expr[:(get_data(S1)[$(idxS1(i, j, m, n))]) for m in 1:dim, n in 1:dim][:]
+        ex2 = Expr[:(get_data(S2)[$(idxS2(m, n, k))])    for m in 1:dim, n in 1:dim][:]
+        push!(exps.args, reducer(ex1, ex2, true))
+    end
+    expr = remove_duplicates(TensorType, exps)
+    quote
+        $(Expr(:meta, :inline))
+        @inbounds return $TensorType($expr)
+    end
+end
```
fredrikekre, line 137: Would be great with a more generic solution to all these things some day. Didn't you start on something Knut?

KnutAM: Not longer than discussing an interface with you. 
With my meta-programming skills, it will take some time. 
Should be relatively easy for someone who understands the meta-programming syntax better though.
My notes for a proposed interface are
```julia
@tensorfun function dcontract(A::Tensor{2,dim}, B::Tensor{2,dim}) where dim    
    C = A[i,j]*B[i,j]
end
@tensorfun function otimes(A::Tensor{2,dim}, B::Tensor{2,dim}) where dim    
    C[i,j,k,l] = A[i,j]*B[k,l]
end
@tensorfun function otimes(A::Tensor{2,2}, B::Tensor{2,2})    
    C[i,j,k,l] = A[i,j]*B[k,l]
end

#=
Restrictions
    * Must be `Tensor{order,dim}` or `SymmetricTensor{order,dim}` (`Vec` not permitted)
    * Output indices sorted alphabetically
=#

# Output, if generic dim given
function \$name(A::Tensor{\$Aorder,dim}, B::Tensor{\$Border,dim}) where dim 
    if dim == 1
        return \$expr_1d
    elseif dim == 2
        return \$expr_2d
    else
        return \$expr_3d
    end
end
# if a specific dim given 
function \$name(A::Tensor{\$Aorder,dim}, B::Tensor{\$Border,dim}) where dim 
    return \$expr
end
```
where some additional decorations should be allowed to specify e.g. SIMD operations and specializations for those types as currently done in Tensors.jl

Another macro would be useful for the AD file too, because this has quite some hard coding and when extending to 3rd order and potentially mixed tensors, this will blow up otherwise.

PR #216, opened 2024-03-15
Hook into Random API

KnutAM:
Following [Hooking into the Random API](https://docs.julialang.org/en/v1/stdlib/Random/#Hooking-into-the-Random-API) I replaced the `Base.rand(Type{<:AbstractTensor})` by `Base.rand(rng::AbstractRNG, ::SamplerType{<:AbstractTensor})`, which has the advantage of supporting

* Using a different generators, e.g. `rand(MersenneTwister(1), Tensor{2,3})`
* Support creating an array of random tensors by e.g. `rand(Vec{2}, 2)`
* Combinations thereof, e.g. `rand(MersenneTwister(2), Vec{2}, 2)`

AFAIU the same API doesn't work for `randn`, so this is kept as is.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -5,17 +5,18 @@ version = "1.16.1"
 [deps]
 ForwardDiff = "f6369f11-7733-5829-9624-2563aa707210"
 LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
-SIMD = "fdea26ae-647d-5447-a871-4b548cad5224"
 PrecompileTools = "aea7be01-6a6a-4083-8856-8a6e6704d82a"
+Random = "9a3f8284-a2c9-5f02-9a11-845980a1fd5c"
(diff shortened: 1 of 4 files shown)


KnutAM:
Ready for review @fredrikekre or @KristofferC

docs/Project.toml
```diff
@@ -1,3 +1,6 @@
 [deps]
 Documenter = "e30172f5-a6a5-5a46-863b-614d45cd2de4"
 Tensors = "48a634ad-e948-5137-8d70-aa71f2a747f4"
+
+[compat]
+Documenter = "0"
```
termi-official:
```suggestion
Documenter = "0"

```

Is this valid?

KnutAM: I guess this was from the pre documenter 1.0 era but was added in automatically, but too long ago to remember if there was a reason... But isn't required since we have the Manifest checked in and could be removed.

KnutAM:
```suggestion
```

PR #218, opened 2024-05-23
Cofactor Matrix

termi-official:
This is commonly used to compute the push-forward of surface areas from reference to current configuration in continuum mechanics (Nansons formula).

```julia
using Tensors, BenchmarkTools

F = Tensor{2,3,Float64}((1.1,-0.1,0.1, -0.1,0.9,0.0 ,0.25,0.0,1.2))

# Master version
cof2(F) = det(F)*transpose(inv(F))
@btime cof2($F); # 16.583 ns
@btime gradient(cof2, $F); # 233.892 ns
@btime gradient(cof2, $F, :all); # 236.603 ns

# This PR
@btime cof($F); # 5.890 ns
@btime gradient(cof, $F); # 78.131 ns
@btime gradient(cof, $F, :all); # 92.096 ns
```

### DIFF
--- a/src/Tensors.jl
+++ b/src/Tensors.jl
@@ -7,7 +7,7 @@ using Statistics: mean
 using LinearAlgebra
 using StaticArrays
 # re-exports from LinearAlgebra
-export ⋅, ×, dot, diagm, tr, det, cof, norm, eigvals, eigvecs, eigen
+export ⋅, ×, dot, diagm, tr, det, norm, eigvals, eigvecs, eigen
 # re-exports from Statistics
 export mean
 
@@ -17,7 +17,7 @@ export otimes, ⊗, ⊡, dcontract, dev, vol, symmetric, skew, minorsymmetric, m
 export otimesu, otimesl
 export minortranspose, majortranspose, isminorsymmetric, ismajorsymmetric
 export tdot, dott, dotdot
-export hessian, gradient, curl, divergence, laplace
+export hessian, gradient, curl, divergence, laplace, cof
 export @implement_gradient
 export basevec, eᵢ
 export rotate, rotation_tensor
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
  - Throw error if trying to call fallback dot method for AbstractArray ([#228])
  - Fix that `curl(f, v::Vec{2})` calls `f(::Vec{3, <:Dual})` instead of `f(::Vec{2, <:Dual})`([#222])
 
+### Added
+ - Optimized version to compute cofactor matrix (also called the minor of a matrix) ([#218])
+
 ## [v1.16.2]
 
 ### Misc
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "Tensors"
 uuid = "48a634ad-e948-5137-8d70-aa71f2a747f4"
-version = "1.16.2"
+version = "1.17.0"
 
 [deps]
 ForwardDiff = "f6369f11-7733-5829-9624-2563aa707210"


fredrikekre:
Can `cof` not be re-used in `inv`? Also, can you add a note in the changelog for 1.17.0 (and update the version number in the project file).

src/Tensors.jl
```diff
@@ -7,7 +7,7 @@ using Statistics: mean
 using LinearAlgebra
 using StaticArrays
 # re-exports from LinearAlgebra
-export ⋅, ×, dot, diagm, tr, det, norm, eigvals, eigvecs, eigen
+export ⋅, ×, dot, diagm, tr, det, cof, norm, eigvals, eigvecs, eigen
```
fredrikekre, line 10: I don't think this come from LinearAlgebra so maybe list separately. Is `cof` a standard name for this operation btw?

termi-official: Oh, right. In the continuum mechanics books I read this operation is usually abbreviated with `cof` (Holzapfel, Wriggers). German Wikipedia also uses it.

https://de.wikipedia.org/wiki/Minor_(Lineare_Algebra)#Eigenschaften

But possible that this abbreviation is a German thing. If you want something different then we can also use it.

termi-official:
> Can `cof` not be re-used in `inv`? Also, can you add a note in the changelog for 1.17.0 (and update the version number in the project file).

We could do, but I am not sure about the performance implications. Will add a benchmark.

src/Tensors.jl
fredrikekre: I don't mind `cof`, just wondering if it was called that, or something else, in other software. I can't find anything from a quick serach though.

termi-official:
If I understood you correctly, then you propose something along this line:
```julia
@generated function inv_new(t::Tensor{2, dim}) where {dim}
    Tt = get_base(t)
    idx(i,j) = compute_index(Tt, i, j)
    ex = quote
        transpose(cof(t))*dinv
    end
    return quote
        $(Expr(:meta, :inline))
        dinv = 1 / det(t)
        @inbounds return $ex
    end
end
```

Seems like this is slightly faster on my machine in 3D.

```julia
julia> F3 = Tensor{2,3,Float64}((1.1,-0.1,0.1, -0.1,0.9,0.0 ,0.25,0.0,1.2))
3×3 Tensor{2, 3, Float64, 9}:
  1.1  -0.1  0.25
 -0.1   0.9  0.0
  0.1   0.0  1.2

julia> F2 = Tensor{2,2,Float64}((1.1,-0.1, -0.1,0.9))
2×2 Tensor{2, 2, Float64, 4}:
  1.1  -0.1
 -0.1   0.9

julia> @btime Tensors.inv_new($F2)
  6.240 ns (0 allocations: 0 bytes)
2×2 Tensor{2, 2, Float64, 4}:
 0.918367  0.102041
 0.102041  1.12245

julia> @btime inv_new($F3)
  11.331 ns (0 allocations: 0 bytes)
3×3 Tensor{2, 3, Float64, 9}:
  0.936281    0.104031    -0.195059
  0.104031    1.12267     -0.0216732
 -0.0780234  -0.00866927   0.849588

julia> @btime Tensors.inv($F2)
  6.240 ns (0 allocations: 0 bytes)
2×2 Tensor{2, 2, Float64, 4}:
 0.918367  0.102041
 0.102041  1.12245

julia> @btime Tensors.inv($F3)
  11.612 ns (0 allocations: 0 bytes)
3×3 Tensor{2, 3, Float64, 9}:
  0.936281    0.104031    -0.195059
  0.104031    1.12267     -0.0216732
 -0.0780234  -0.00866927   0.849588
```

Reproducer
```julia
using Tensors, BenchmarkTools

F2 = Tensor{2,2,Float64}((1.1,-0.1, -0.1,0.9))
F3 = Tensor{2,3,Float64}((1.1,-0.1,0.1, -0.1,0.9,0.0 ,0.25,0.0,1.2))

@btime Tensors.inv_new($F2)
@btime Tensors.inv_new($F3)
@btime Tensors.inv($F2)
@btime Tensors.inv($F3)
```

Should I include the change?

KnutAM:
I'm getting 
```julia
julia> @btime inv_new($F2);
  2.100 ns (0 allocations: 0 bytes)

julia> @btime inv_new($F3);
  6.800 ns (0 allocations: 0 bytes)

julia> @btime Tensors.inv($F2);
  1.900 ns (0 allocations: 0 bytes)

julia> @btime Tensors.inv($F3);
  4.300 ns (0 allocations: 0 bytes)
```

with 
```julia
@generated function inv_new(t::Tensor{2, dim}) where {dim}
           Tt = Tensors.get_base(t)
           idx(i,j) = Tensors.compute_index(Tt, i, j)
           ex = quote
               transpose(cof(t))*dinv
           end
           return quote
               $(Expr(:meta, :inline))
               dinv = 1 / det(t)
               @inbounds return $ex
           end
       end
end
```

KnutAM:
Could do
```julia
cof(t::SecondOrderTensor) = _cof(t, Val(false))
inv(t::SecondOrderTensor) = _cof(t, Val(true)) / det(t)
@generated function _cof(t::Tensor{2, dim, T}, ::Val{Transpose}) where {dim, T, Transpose}
    Tt = get_base(t)
    idx(i,j) = Transpose ? compute_index(Tt, j, i) : compute_index(Tt, i, j)
    ...
```
but not sure if that is worth the added complexity compared to just duplicating the code...

src/Tensors.jl
termi-official: My only worry here is that some standard library might include cofactors in the future. :D

PR #233, opened 2025-12-26
Use `get_expression` in generated functions

KnutAM:
As an alternative to #211 as I'm not content with the macro code, this only introduces the expression generator, but not the macro expression parsing which gave a lot of code which I found look hard to maintain/understand when looking back. 

Todo’s
- [x] Use in #188
- [x] Create a benchmark script to validate performance for different platforms

~Note that the `tensor_product` macro builds on this function, so that higher level API can be added later~
Edit: Not anymore, but should be possible if desired. 

### Benchmark summary
Apple M3, Intel i9, MD Ryzen 9 all give similar timings

PR #236, opened 2026-01-04
MixedTensor

KnutAM:
Supersedes #188 as that got closed when merging #233

TODOs

* [x] Reparameterize into using `Tuple{d1, d2, ...}` to allow better dispatch control
* [x] Add better test coverage for mixed cases 
* [x] Validate that no performance regression occurs due to generalization of AD
* [x] Open PR in Ferrite.jl to use `MixedTensor` in embedded cases (https://github.com/Ferrite-FEM/Ferrite.jl/pull/1286)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [v1.17.0]
+### Added
+
+- Support for `MixedTensors` and enhanced support for 3rd order tensors ([#236])
+- Internal code-generation generalization for tensor products ([#233])
 
 ### Bugfixes
  
@@ -43,4 +47,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 [v1.16.2]: https://github.com/Ferrite-FEM/Tensors.jl/releases/tag/v1.16.2
 [#205]: https://github.com/Ferrite-FEM/Tensors.jl/issues/205
 [#212]: https://github.com/Ferrite-FEM/Tensors.jl/issues/212
+[#222]: https://github.com/Ferrite-FEM/Tensors.jl/issues/222
 [#225]: https://github.com/Ferrite-FEM/Tensors.jl/issues/225
+[#228]: https://github.com/Ferrite-FEM/Tensors.jl/issues/228
+[#233]: https://github.com/Ferrite-FEM/Tensors.jl/issues/233
+[#236]: https://github.com/Ferrite-FEM/Tensors.jl/issues/236


src/Tensors.jl
```diff
@@ -66,6 +67,15 @@ struct Tensor{order, dim, T, M} <: AbstractTensor{order, dim, T}
     Tensor{order, dim, T, M}(data::NTuple) where {order, dim, T, M} = new{order, dim, T, M}(data)
 end
 
+struct MixedTensor{order, dims, T, M} <: AbstractTensor{order, dims, T}
```
termi-official: A docstring will be of help for users (as well as an entry in the docs).

src/automatic_differentiation.jl
```diff
@@ -177,6 +189,23 @@ end
...
+    @inbounds begin
+        p = map(partials, get_data(v)) # tuple((partials(vi) for vi in get_data(v))...)
+        N1 = n_components(MixedTensor{2,dims1})
+        N2 = n_components(MixedTensor{2,dims2})
+        ∇f = MixedTensor{4, (dims1[1],dims1[2],dims2[1],dims2[2])}(
+            ntuple(i->p[rem(i-1,N1)+1][div(i-1,N1)+1], N1*N2)
+        )
+    end
+    return makeregular(∇f)
+end
```
termi-official: Can we make this generic, or at least add the cases for MixedTensor by Vec and Vec by MixedTensor? Also, is the `makeregular` here correct?

KnutAM:
> Any chance that the implementation of product operations got lost? I cannot find single contraction, double contraction and the outer products.

should work with the same code as for regular tensors via `get_expression`

src/Tensors.jl
```diff
@@ -76,8 +86,8 @@ const AllTensors{dim, T} = Union{SymmetricTensor{2, dim, T}, Tensor{2, dim, T},
                                  Vec{dim, T}, Tensor{3, dim, T}}
 
 
-const SecondOrderTensor{dim, T}   = Union{SymmetricTensor{2, dim, T}, Tensor{2, dim, T}}
-const FourthOrderTensor{dim, T}   = Union{SymmetricTensor{4, dim, T}, Tensor{4, dim, T}}
+const SecondOrderTensor{dim, T}   = Union{SymmetricTensor{2, dim, T}, Tensor{2, dim, T}, MixedTensor{2, dim, T}}
+const FourthOrderTensor{dim, T}   = Union{SymmetricTensor{4, dim, T}, Tensor{4, dim, T}, MixedTensor{4, dim, T}}
```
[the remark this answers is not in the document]
termi-official: I am not sure about how these are used downstream, so I wanted to make sure we are at least aware that there might be potential issues.

KnutAM:
When implementing in [Ferrite](https://github.com/Ferrite-FEM/Ferrite.jl/pull/1286), I noticed that the parameterization is a bit annoying for dispatch, since 

`foo(t::MixedTensor{2, (dim, 1)}) where {dim}` is not possible.

An alternative (like `SArray`), is having e.g. 
```julia
struct MixedTensor{order, dims <: Tuple, T, M} <: AbstractTensor{order, dims, T}
    data::NTuple{M, T}
end
const MixedTensor2{d1, d2, T, M} = MixedTensor{2, Tuple{d1, d2}, T, M}
const MixedTensor3{d1, d2, d3, T, M} = MixedTensor{3, Tuple{d1, d2, d3}, T, M}
const MixedTensor4{d1, d2, d3, d4, T, M} = MixedTensor{4, Tuple{d1, d2, d3, d4}, T, M}
```
as this would allow `foo(t::MixedTensor2{dim, 1}) where {dim}`. 

I was thinking that it should be able to make a similar constructor as the current version with this too (like `MixedTensor{2, (1,2)}(data)`), but can't get the dispatches right with
```julia
function MixedTensor{order, dims, T, M}(data::NTuple{M, T}) where {order, dims, T, M}
    return MixedTensor{order, Tuple{dims...}, T, M}(data)
end
```
...

KnutAM:
Benchmark test shows no regression on AD and good performance for MixedTensors. 
```julia
using Tensors, BenchmarkTools
foo(x) = x * norm(x ⋅ x')
bar(x) = gradient(foo, x)

# master
julia> @btime bar($(rand(Tensor{2,3})));
  145.113 ns (0 allocations: 0 bytes)

julia> @btime gradient(norm, $(rand(Tensor{2,3})));
  33.568 ns (0 allocations: 0 bytes)

julia> @btime gradient(identity, $(rand(Tensor{2,3})));
  10.541 ns (0 allocations: 0 bytes)

# PR
julia> @btime bar($(rand(Tensor{2,3})));
  145.362 ns (0 allocations: 0 bytes)

julia> @btime gradient(norm, $(rand(Tensor{2,3})));
  33.652 ns (0 allocations: 0 bytes)

julia> @btime gradient(identity, $(rand(Tensor{2,3})));
  10.541 ns (0 allocations: 0 bytes)
```

And if we use a mixed tensor-case, we have e.g. 
```julia
julia> @btime bar($(rand(MixedTensor2{2,3})));
  50.700 ns (0 allocations: 0 bytes)

julia> @btime bar($(rand(MixedTensor2{3,2})));
  77.203 ns (0 allocations: 0 bytes)

julia> @btime gradient(norm, $(rand(MixedTensor2{2,3})));
  15.071 ns (0 allocations: 0 bytes)

julia> @btime gradient(identity, $(rand(MixedTensor2{2,3})));
  4.166 ns (0 allocations: 0 bytes)
```

Could make the benchmark suite run properly. Fixed one error (`eig` instead of `eigen`), but got more I didn't know what went wrong with. Fixing that is outside the scope of this PR, so leaving that untouched.

PR #244, opened 2026-04-07
Optimized Moore-Penrose psuedoinverse

termi-official:
Closes #241

### DIFF
--- a/src/math_ops.jl
+++ b/src/math_ops.jl
@@ -201,19 +201,37 @@ julia> pinv(A)
   0.665941  -0.179303   0.787808
 ```
 """
-function LinearAlgebra.pinv(A::Tensor{2})
-    return inv(tdot(A)) ⋅ A'
-end
-function LinearAlgebra.pinv(A::SymmetricTensor{2})
-    return unsafe_symmetric(inv(tdot(A)) ⋅ A')
-end
-@generated function LinearAlgebra.pinv(A::MixedTensor2{d1, d2}) where {d1, d2}
-    ex = if d1 > d2
-        quote inv(tdot(A)) ⋅ (A)' end
-    else
-        quote (A)' ⋅ inv(dott(A)) end
+@inline function LinearAlgebra.pinv(A::AbstractTensor{2})
+    tdot_A_or_dott_A = _tdot_or_dott(A) # Symmetric
+    det_tol = eps(norm(tdot_A_or_dott_A)^size(tdot_A_or_dott_A, 1))
+    if abs(det(tdot_A_or_dott_A)) > det_tol # tdot_A_or_dott_A invertible
+        return _fast_pinv(A, tdot_A_or_dott_A)
+    else # If not invertible, use generic (slow) fallback via StaticArrays)
+        return _generic_pinv(A)
     end
-    return ex
+end
+@inline _tdot_or_dott(A::Union{Tensor{2}, SymmetricTensor{2}}) = tdot(A)
+@inline function _tdot_or_dott(A::MixedTensor2{d1, d2}) where {d1, d2}
+    return d1 > d2 ? tdot(A) : dott(A)
+end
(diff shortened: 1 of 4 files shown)


src/math_ops.jl
```diff
@@ -182,6 +182,40 @@ end
...
+    return unsafe_symmetric(inv(tdot(A)) ⋅ A')
+end
+@generated function LinearAlgebra.pinv(A::MixedTensor2{d1, d2}) where {d1, d2}
+    ex = if d1 > d2
+        quote inv(tdot(A)) ⋅ (A)' end
+    else
+        quote (A)' ⋅ inv(dott(A)) end
+    end
+    return ex
+end
```
KnutAM: To make this work for more tricky cases, such as 
```julia
x = Tensor{2,2}((1.0, 1.0, 0.0, 0.0))
x = Tensor{2,2}((1.0, 1.0, 1.0, 1.0))
x = zero(MixedTensor2{3,2})
```
I suggest adding a generic fallback, something like

```suggestion
function LinearAlgebra.pinv(A::AbstractTensor{2})
    tdot_A_or_dott_A = _tdot_or_dott(A) # Symmetric
    det_tol = eps(norm(tdot_A_or_dott_A)^size(tdot_A_or_dott_A, 1))
    if abs(det(tdot_A_or_dott_A)) > det_tol # tdot_A_or_dott_A invertible
        return _fast_pinv(A, tdot_A_or_dott_A)
    else # If not invertible, use generic (slow) fallback via StaticArrays)
        return _generic_pinv(A)
    end
end

@inline _tdot_or_dott(A::Union{Tensor{2}, SymmetricTensor{2}}) = tdot(A)
@inline function _tdot_or_dott(A::MixedTensor2{d1, d2}) where {d1, d2}
    return d1 > d2 ? tdot(A) : dott(A)
end

function _generic_pinv(A::Tensor{2, dim}) where {dim}
    return Tensor{2, dim}(pinv(SMatrix{dim, dim}(get_data(A))).data)
end
function _generic_pinv(A::SymmetricTensor{2, dim}) where {dim}
    return SymmetricTensor{2, dim}(pinv(Symmetric(SMatrix{dim, dim}(A))))
end
function _generic_pinv(A::MixedTensor2{d1, d2}) where {d1, d2}
    return MixedTensor2{d2, d1}(pinv(SMatrix{d1, d2}(get_data(A))).data)
end

@inline _fast_pinv(A::Tensor{2}, tdot_A) = inv(tdot_A) ⋅ A'
@inline _fast_pinv(A::SymmetricTensor{2}, tdot_A) = unsafe_symmetric(inv(tdot_A) ⋅ A')

@inline function _fast_pinv(A::MixedTensor2{d1, d2}, tdot_A_or_dott_A) where {d1, d2}
    if d1 > d2
        return inv(tdot_A_or_dott_A) ⋅ (A)'
    else
        return (A)' ⋅ inv(tdot_A_or_dott_A)
    end
end
```

KnutAM (changes requested):
Nice!

I don't think we need generated functions here to check `d1>d2`, since that should be const-propped away? Need to validate with benchmarks to be sure, but at least reasonably fast this way too. 

Needs also tests for the more complicated cases.

src/math_ops.jl
termi-official: This code is 3x slower on my machine and has an allocation:

This code
```julia
julia> A = rand(MixedTensor2{3,2})
3×2 MixedTensor2{3, 2, Float64, 6}:
 0.646457  0.502131
 0.363409  0.161386
 0.877468  0.256305

julia> @btime pinv($A)
  42.505 ns (1 allocation: 64 bytes)
```

The code above
```julia
julia> A = rand(MixedTensor2{3,2})
3×2 MixedTensor2{3, 2, Float64, 6}:
 0.292221  0.895357
 0.104659  0.727678
 0.943476  0.390033

julia> @btime pinv($A)
  14.970 ns (0 allocations: 0 bytes)
```

